PR #18, opened 2026-01-15
codereview: require real-behavior tests for new components

xingyaoww:
## Summary
- Updates `/codereview` and `/codereview-roasted` workflows to require corresponding tests when new components/behaviors are introduced (when the repo has test infrastructure).
- Adds explicit guidance that tests should exercise real behavior (not just mock-call assertions).

## Changes
- `skills/codereview/SKILL.md`: adds a "Testing and Behavior Verification" review scenario + example.
- `skills/codereview-roasted/SKILL.md`: adds a "Testing and Regression Proof" scenario + "[TESTING GAPS]" output section.

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

### DIFF
--- a/skills/codereview/SKILL.md
+++ b/skills/codereview/SKILL.md
@@ -44,7 +44,8 @@ Watch for:
 If the repository has a test infrastructure (unit/integration/e2e tests) and the PR introduces new components, modules, routes, CLI commands, user-facing behaviors, or bug fixes, ensure there are corresponding tests.
 
 When reviewing tests, prioritize tests that validate real behavior over tests that primarily assert on mocks:
-- Prefer exercising real code paths (e.g., parsing, validation, business logic), using in-memory or lightweight fakes only where necessary (e.g., ephemeral DB, temp filesystem).
+- Prefer tests that exercise real code paths (e.g., parsing, validation, business logic) and assert on outputs/state.
+- Use in-memory or lightweight fakes only where necessary (e.g., ephemeral DB, temp filesystem) to keep tests fast and deterministic.
 - Flag tests that only mock the unit under test and assert it was called, unless they cover a real coverage gap that cannot be achieved otherwise.
 - Ensure tests fail for the right reasons (i.e., would catch a regression), and are not tautologies.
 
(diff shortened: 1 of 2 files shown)


skills/codereview/SKILL.md
```diff
@@ -40,6 +40,14 @@ Watch for:
 - Incorrect use of cryptographic libraries
 - Common pitfalls (null dereferencing, off-by-one errors, race conditions)
 
+4. Testing and Behavior Verification
+If the repository has a test infrastructure (unit/integration/e2e tests) and the PR introduces new components, modules, routes, CLI commands, user-facing behaviors, or bug fixes, ensure there are corresponding tests.
+
+When reviewing tests, prioritize tests that validate real behavior over tests that primarily assert on mocks:
+- Avoid tests that only mock the unit under test and assert it was called.
```
enyst: I feel GPT-5.2 will really pay attention to this 😅 

Yes, it's fair, and yet sometimes those are useful. I think I agree though, we really need tests actually exercising the paths, not a maze of mocks that, after 3+ anyway, become unreadable to humans so... hmm. (this might be particularly true about the servers I feel)

Maybe we could say like
> Prefer exercising real code paths (e.g., parsing, validation, business logic), using in-memory or lightweight fakes only where necessary (e.g., ephemeral DB, temp filesystem). Flag tests that only mock the unit under test and assert it was called, unless they cover a real coverage gap that cannot be achieved otherwise.

🤔 I could be wrong though, maybe we should just give those up.

skills/codereview/SKILL.md
```diff
@@ -52,5 +60,7 @@ Use the following structure in your output:
 [src/utils.py, Line 42] :hammer_and_wrench: Unused import: The 'os' module is imported but never used. Remove it to clean up the code.
 [src/database.py, Lines 78–85] :mag: Readability: This nested if-else block is hard to follow. Consider refactoring into smaller functions or using early returns.
 [src/auth.py, Line 102] :closed_lock_with_key: Security Risk: User input is directly concatenated into an SQL query. This could allow SQL injection. Use parameterized queries instead.
+[tests/test_auth.py, Lines 12–45] :test_tube: Testing: This PR adds new behavior but the tests only assert mocked calls. Add a test that exercises the real code path and asserts on outputs/state so it would catch regressions.
```
enyst, line 64: Got it, yes, this is fair.

skills/codereview/SKILL.md
xingyaoww: @OpenHands please apply engel's suggestion here and modify the prompt

enyst (approved):
Thank you! This will help address a problem I think about for a while, I do think it's real and it may become worse in time as we fill in coverage with mocks...

PR #31, opened 2026-02-06
'add-skill' Agent Skill

jpelletier1:
Introduces an `/add-skill` Agent Skill for copying an Agent Skill from any repo and installing it into the `.openhands/skills` directory within the current conversation workspace. Also makes the Agent Skill ready for use.

Usage: `/add-skill https://github.com/OpenHands/skills/tree/main/skills/codereview`


enyst (approved):
Hah, thank you! Funny one. I think the agent can anyway, but definitely it may help it if it works smoothly. (and its human!)

Let’s merge and test it live.

PR #91, opened 2026-03-05
Add consideration for external test costs in code review skills

### LINKED ISSUE #90: PR review skills should take into account cost of doing external tests
Some PRs come with a proof of manual test or some test outside the test suite.
This tests have a non negligible cost, someone took the time to do them and review that the test was correct.

Any change to the PR after would require redoing the tests or risk merging untested work.
In these cases, all suggestions should consider if they are important enough to redo the tests.

juanmichelini:
## Summary

This PR adds a minimal single sentence to both code review skills (`code-review` and `codereview-roasted`) to address the cost of external or manual tests when reviewing PRs.

## Changes

Added the following guidance after the TASK section in both skills:
> Before suggesting changes, check if the PR includes manual or external test results, and if so, weigh whether each suggestion is important enough to warrant re-testing.

## Rationale

Some PRs include manual tests or tests outside the test suite that have a non-negligible cost. Any change to the PR after these tests would require redoing them or risk merging untested work. This addition ensures reviewers consider whether suggestions are important enough to justify that cost.

Fixes #90

### DIFF
--- a/skills/code-review/SKILL.md
+++ b/skills/code-review/SKILL.md
@@ -11,6 +11,8 @@ You are an expert software engineer and code reviewer with deep experience in mo
 TASK:
 Review the code changes in this pull request or merge request, and provide actionable feedback on **important issues only**. Focus on bugs, security, and correctness - skip minor style nits. If the code is good, just approve it. DO NOT modify the code; only provide specific feedback.
 
+Before suggesting changes, check if the PR includes manual or external test results, and if so, weigh whether each suggestion is important enough to warrant re-testing.
+
 CONTEXT:
 You have full context of the code being committed in the pull request or merge request, including the diff, surrounding files, and project structure. The code is written in a modern language and follows typical idioms and patterns for that language.
 
--- a/skills/codereview-roasted/SKILL.md
+++ b/skills/codereview-roasted/SKILL.md
@@ -25,6 +25,8 @@ Before reviewing, ask Linus's Three Questions:
 TASK:
 Provide brutally honest, technically rigorous feedback on code changes. Be direct and critical while remaining constructive. Focus on fundamental engineering principles over style preferences. DO NOT modify the code; only provide specific, actionable feedback.
 
+Before suggesting changes, check if the PR includes manual or external test results, and if so, weigh whether each suggestion is important enough to warrant re-testing.
+
 CODE REVIEW SCENARIOS:
 
 1. **Data Structure Analysis** (Highest Priority)


enyst:
Please correct me if I’m wrong, the reason for not saying stuff here would be to make it APPROVE, rather than COMMENT, when there were live tests/evals already made, do I understand correctly? 

I see the point of not redoing those tests, and the reviewer agent surely can be annoying, hah. But what I wonder is: if that is the case, is it a case where we should just ask for human review, rather than telling the agent to stop posting comments after a certain stage, WDYT?

juanmichelini:
@enyst, yeah it is a fine line.

the case I'm thinking is:

Dev post PR with external costly tests.
Agent says "acceptable but hey make this more readable"
Dev can choose between:
a) calling a human to review it,
b) accepting the change of the agent and redoing tests manually
c) thinking the change is innocuous enough and merge it without doing tests again

my thoughts: 
a) is okay, but if it happens often then the bot is not doing its job.
b) sounds like the right choice, but if the changes are nits it's actually moving eyeballs in the wrong direction.
c) is as likely to break things as it is too help.

juanmichelini:
Maybe related https://github.com/OpenHands/extensions/issues/92

enyst:
Good points! A few thoughts about this:

> Dev can choose between:
a) calling a human to review it,
b) accepting the change of the agent and redoing tests manually
c) thinking the change is innocuous enough and merge it without doing tests again

I think the question underlying this choice is a question of risk: how risky is the PR overall, and how risky are the changes?

Initially, I remember Xingyao had a prompt for the reviewer agent when it was relatively restricting it to _"straightforward and low-risk"_ PRs, e.g. change model config, or CI: [initial prompt](https://github.com/OpenHands/software-agent-sdk/commit/519ebae0c353e644110371c1196002717c76c450#diff-85c4ddc2dc26f8d29524f4a305f2c695bd6627fe43ed68c2f99cbc6236e7606dR18)

All the rest, would remain under human review.

The funniest thing happened next, like, dunno, within days: Xingyao got a bunch reviews that were clearly saying that it's all good, but the agent wouldn't approve, so he "convinced" it 😅 😅 😅  
[Slack fun](https://openhands-ai.slack.com/archives/C06PB3T5ZK6/p1770832251114859)

And we changed the prompt 😂 . Now it's "_Default to APPROVE_" and several other pieces: [policy](https://github.com/OpenHands/software-agent-sdk/blob/main/.agents/skills/custom-codereview-guide.md#default-approval-policy)

---

I think, which brings us to your second comment, actually it makes a lot of sense to default to human review, like, basically risky enough or within the area of interest of a human.

My big question on this:
> c) thinking the change is innocuous enough and merge it without doing tests again
> c) is as likely to break things as it is too help.

is how bad is the breaking things part?
- is it likely to be a one-line fix? Surely not redo manual tests. 
- is it likely to be some easy / fast fix? Not redo manual tests. 
- does it touch multiple components in the system? 🤔 Redo manual tests or ask a human directly or both.
- does it touch deeply one component? Ask the human who's been working more on that before / cares about that.
- is it a refactoring, 4+ files? Ask a human, probably...ish.

PR #114, opened 2026-03-19
feat: add agentskill-learn skill for 110K+ skill discovery

romainsimon:
## What

Adds an `agentskill-learn` skill that connects OpenHands to the [agentskill.sh](https://agentskill.sh) registry, giving users access to 110,000+ cross-platform AI agent skills.

## Why

OpenHands users currently need to manually find and import skills. This skill adds a built-in discovery layer that lets agents search, install, and manage skills from the largest open skills registry.

The skill is a lightweight bootstrap (~65 lines) that fetches the full `/learn` skill from [agentskill-sh/learn](https://github.com/agentskill-sh/learn) on first use via the existing `add-skill` mechanism. This means:

- **No maintenance burden**: the bootstrap is stable and rarely changes
- **Always up-to-date**: the real skill is fetched fresh from GitHub
- **No duplication**: skill logic lives in one canonical repo

## Changes

- `skills/agentskill-learn/SKILL.md`: Bootstrap skill with quick-start API docs and setup instructions
- `marketplaces/default.json`: Added marketplace entry under "productivity" category

## How Tested

- Verified SKILL.md follows the OpenHands skill format (YAML frontmatter with name, description, triggers)
- Verified marketplace entry follows existing patterns (source, category, keywords)
- API endpoints tested: search, install, and trending all return valid JSON

### DIFF
--- a/marketplaces/default.json
+++ b/marketplaces/default.json
@@ -23,6 +23,20 @@
             ]
         },
         {
+            "name": "agentskill-learn",
+            "source": "./agentskill-learn",
+            "description": "Discover, install, and manage 110,000+ AI agent skills from agentskill.sh. Search by keyword, install mid-session, scan for security issues, and rate skills after use.",
+            "category": "productivity",
+            "keywords": [
+                "skills",
+                "marketplace",
+                "registry",
+                "install",
+                "discover",
+                "agentskill"
+            ]
+        },
+        {
             "name": "agent-memory",
             "source": "./agent-memory",
             "description": "Persist and retrieve repository-specific knowledge using AGENTS.md files. Use when you want to save important information about a codebase (build commands, code style, workflows) for future sessions.",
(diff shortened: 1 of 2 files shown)


enyst (commented):
Hey @romainsimon, thank you for this !

I think we should connect agentskills.sh, I have a two little questions:
- I see that the learn skill uses `AskUserQuestion`, which afaik is a Claude Code tool. OpenHands doesn't define such tool. We have a way to "ask question" the other way around: the user can ask a question about the conversation which is outside agent's history, and I don't think that helps here. 😅 WDYT we could do here, adapt the `learn` skill to other platforms or would you propose the inclusion of some similar `AskUserQuestion` in OH? ([example](https://github.com/agentskill-sh/learn/blob/main/SKILL.md#output-formatting-guidelines))
- we are trying to offer the ability for the user to define multiple marketplaces, and in this repo I believe we have already defined more. Could defining the marketplace or something like that be an alternative way to give the agent the ability to use agentskills.sh?

PR #118, opened 2026-03-23
feat(vulnerability-remediation): support upgrading transitive dependencies without adding as top-level

aivong-openhands:
## Summary

Port changes from https://github.com/OpenHands/vulnerability-fixer/pull/48

When a CVE fix requires upgrading a transitive dependency (not directly declared in manifest files), the remediation agent now attempts to upgrade it **without adding it as a top-level dependency** in the manifest file.

### `scan_and_remediate.py`
- Added Step 2: Determine if direct or transitive dependency
- Added Step 2a: Upgrading TRANSITIVE Dependencies with package manager-specific commands (uv, Poetry, npm, yarn, pnpm, Cargo, Go)
- Updated UV lockfile section with transitive upgrade guidance
- Renumbered subsequent steps

### `README.md`
- Updated workflow steps to include dependency type determination
- Added "Direct vs Transitive Dependencies" section
- Added "Transitive Dependency Upgrades" section with package manager commands

## Package Manager Commands for Transitive Upgrades

| Package Manager | Command |
|-----------------|---------|
| uv | `uv lock --upgrade-package <package>` |
| Poetry | `poetry update <package>` |
| npm | `npm update <package>` or `overrides` |
| yarn | `yarn upgrade <package>` or `resolutions` |
| pnpm | `pnpm update <package>` or `overrides` |
| Cargo | `cargo update -p <package>` |
| Go | `go get <package>@v<version>` + `go mod tidy` |

## References

- https://github.com/OpenHands/vulnerability-fixer/pull/48

### DIFF
--- a/plugins/vulnerability-remediation/README.md
+++ b/plugins/vulnerability-remediation/README.md
@@ -146,15 +146,44 @@ You can also trigger scans manually:
 When remediating a vulnerability, the agent:
 
 1. **Analyzes** the vulnerability details (CVE ID, affected package, versions)
-2. **Locates** the dependency file (package.json, requirements.txt, pom.xml, etc.)
-3. **Updates** the package to the fixed version
-4. **Verifies** the change doesn't break the build
-5. **Creates a branch** named `fix/<cve-id>`
-6. **Commits** changes with a descriptive message
-7. **Creates a PR** with:
+2. **Determines** if the package is a direct or transitive dependency
+3. **For transitive dependencies**: Attempts lockfile-only upgrade without adding to manifest
+4. **For direct dependencies**: Updates the package version in manifest files
+5. **Regenerates** all corresponding lockfiles
+6. **Verifies** the change doesn't break the build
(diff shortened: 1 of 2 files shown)


neubig (commented):
Hi @aivong-openhands I kinda agree with the auto-reviewer that the prompt is becoming very complex. Do we need all of this info? If so maybe it's OK, but if not we could simplify.

Further to consider: maybe we could refactor this repo to rely on the skill in the OpenHands/extensions repo? Then OpenHands/extensions could be a single source of truth for these directions. This doesn't necessarily need to be addressed in this PR though.

juanmichelini:
@aivong-openhands  I'm doing some housekeeping, is this still ongoing?

aivong-openhands:
Closing as we haven't need this 🤔 I think openhands via the vulnerability fixer has done an OK job of updating transitive dependencies. Can create a new PR if the issue resurfaces

PR #130, opened 2026-04-01
fix(pr-review): skip installing project dependencies during review

Shimada666:
## Summary

- Add `--no-project` flag to `uv run` in the PR review action so it **only installs the three required packages** (`openhands-sdk`, `openhands-tools`, `lmnr`) instead of all dependencies from the reviewed project's `pyproject.toml`.

## Problem

When the action runs `cd pr-repo && uv run --with ...`, `uv` detects the project's `pyproject.toml` in the current directory and installs **all** project dependencies first. For projects with heavy deps (e.g. opencv, mediapipe, torch), this adds significant time and can even fail if native build tools are missing.

## Fix

```diff
- uv run --with openhands-sdk --with openhands-tools --with lmnr \
+ uv run --no-project --with openhands-sdk --with openhands-tools --with lmnr \
```

The review agent only reads source files and calls the LLM API — it never imports or runs the project code, so project dependencies are unnecessary. The `--no-project` flag tells `uv` to ignore the local `pyproject.toml` while still keeping `pr-repo/` as the working directory for file access.

Ref: https://docs.astral.sh/uv/reference/cli/#uv-run--no-project

### DIFF
--- a/plugins/pr-review/action.yml
+++ b/plugins/pr-review/action.yml
@@ -133,7 +133,7 @@ runs:
               REPO_NAME: ${{ github.repository }}
           run: |
               cd pr-repo
-              uv run --with openhands-sdk --with openhands-tools --with lmnr \
+              uv run --no-project --with openhands-sdk --with openhands-tools --with lmnr \
                 python ../extensions/plugins/pr-review/scripts/agent_script.py
 
         - name: Upload logs as artifact


xingyaoww:
@OpenHands /codereview-roasted /github-pr-review

plugins/pr-review/action.yml
```diff
@@ -133,7 +133,7 @@ runs:
               REPO_NAME: ${{ github.repository }}
           run: |
               cd pr-repo
-              uv run --with openhands-sdk --with openhands-tools --with lmnr \
+              uv run --no-project --with openhands-sdk --with openhands-tools --with lmnr \
```
xingyaoww, line 136: 🟢 **Acceptable**: Clean, minimal, correct. `--no-project` is positioned before `--with` which reads naturally ("ignore the project, but give me these packages"). The `uv` docs confirm this is the intended usage. Nothing to add.

xingyaoww (approved):
🟢 **Good taste** — This is exactly the kind of one-line fix that separates people who understand their tools from people who just cargo-cult commands.

**Linus-Style Analysis:**

The problem was real: `uv run` in a directory with a `pyproject.toml` blindly resolves the entire dependency tree of a project the review agent will *never import*. For projects with heavy native deps (opencv, torch, mediapipe), this means minutes of wasted CI time or outright build failures because some C library isn't installed. That's not a theoretical concern — that's a production bug.

The fix is one flag: `--no-project`. It tells `uv` to stop being "helpful" and just install what we actually asked for. The working directory stays the same, file access is preserved, the three packages we need (`openhands-sdk`, `openhands-tools`, `lmnr`) still get installed. Nothing breaks, everything gets faster.

This is what good engineering looks like. You identified the root cause, applied the minimal correct fix, and didn't bolt on some Rube Goldberg workaround. No wrapper scripts, no conditional logic, no "let's cache the deps" band-aid — just the right flag for the right tool.

**Verdict:** ✅ Ship it.

**Key Insight:** The review agent reads files and talks to an LLM — it has no business installing the reviewed project's dependency tree.

---
_This review was generated by an AI agent (OpenHands) on behalf of the user._

PR #180, opened 2026-04-15
plugins/pr-review: add opt-in enable-uv-cache input

09473ZH:
## What

Adds `enable-uv-cache` input (default `false`, wired to `astral-sh/setup-uv@v6`'s `enable-cache`). Default behavior is unchanged.

## Why

Every PR review currently re-downloads OpenHands SDK + its transitive deps. Measured on a self-hosted runner: **~500 MB disk writes + ~300 MB PyPI traffic per review**, concentrated in 20-30 second spikes that saturate single SSDs when multiple runners trigger concurrently.

The existing comment explains the reason caching is off:

> shared caches can become a pivot into more privileged workflows

That's a valid concern for hosted runners and any repo where the review workflow and higher-privilege workflows share cache storage — a prompt-injected reviewer could write a malicious wheel that a later workflow consumes. **The default should stay `false`.**

Single-tenant self-hosted runners don't cross that trust boundary. Both the reviewer and any other consumer of the cache are operated by the same team. For those users, opting into caching cuts the per-run overhead substantially with no new attack surface.

## Changes

- `action.yml`: new `enable-uv-cache` input (default `'false'`), threaded to `setup-uv`'s `enable-cache`
- `README.md`: new "Caching and Security" section explaining the threat model, when to opt in, and an alternative for self-hosted runners (host-level uv cache volume via Docker)

## Backward compat

Default is unchanged — existing users see no behavior difference until they explicitly set `enable-uv-cache: 'true'`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/pr-review/README.md
+++ b/plugins/pr-review/README.md
@@ -146,6 +146,28 @@ PR reviews are automatically triggered when:
 | `llm-api-key` | Yes | - | LLM API key |
 | `github-token` | Yes | - | GitHub token for API access |
 | `lmnr-api-key` | No | `''` | Laminar API key for observability |
+| `enable-uv-cache` | No | `'false'` | Enable setup-uv's GitHub Actions cache for Python deps. Default `false` for security (see [Caching and Security](#caching-and-security)). |
+
+## Caching and Security
+
+Python dependency caching is **disabled by default**. `uv run --with ...` re-downloads OpenHands SDK and its transitive deps on every run, which is slow but safe.
+
+**Why it's off by default:** Prompt injection can coerce the reviewer into executing arbitrary commands during the review. A compromised review run could write a malicious wheel into the shared GitHub Actions cache. Any later, higher-privilege workflow in the same repository that hits the same cache key would silently execute the attacker's code — a supply-chain pivot.
+
+**Enabling it is safe when:**
+- The runner is single-tenant (e.g. your own self-hosted runner, not shared with untrusted workflows).
+- You do not run other privileged workflows in the same repository that would consume setup-uv's cache.
+- You accept the residual risk in exchange for faster runs / lower disk writes.
+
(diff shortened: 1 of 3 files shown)


plugins/pr-review/README.md
```diff
@@ -146,6 +146,28 @@ PR reviews are automatically triggered when:
...
+- You accept the residual risk in exchange for faster runs / lower disk writes.
+
+**Self-hosted runners:** Consider mounting a host-level uv cache volume (e.g. `/home/runner/.cache` as a Docker volume) instead of — or in addition to — this option. A local volume is faster than a round trip to GHA cache storage and does not cross any trust boundary.
+
+```yaml
+- uses: OpenHands/extensions/plugins/pr-review@main
+  with:
+    enable-uv-cache: true   # opt in on trusted self-hosted runners
+    llm-api-key: ${{ secrets.LLM_API_KEY }}
+    github-token: ${{ secrets.CORGI_BOT_GITHUB_PAT }}
```
enyst: I don’t think this was intentional? 😅

enyst (changes requested):
Thank you, these are interesting points! Left a small comment inline 🤔

plugins/pr-review/README.md
```diff
@@ -146,6 +146,28 @@ PR reviews are automatically triggered when:
...
+```
```
xingyaoww:
```suggestion
```

xingyaoww (approved):
LGTM after removing that section @enyst mentioned

plugins/pr-review/README.md
09473ZH: No it wasn't intentional — accidentally pulled in our internal bot PAT name (`CORGI_BOT_GITHUB_PAT`) when I copied an example from our downstream workflow. Sorry about that and thanks for the catch! Looks like @xingyaoww has already removed that block in fe4445b, so it should be addressed. Could you take another look? 🙏

enyst (approved):
Thank you!

PR #193, opened 2026-04-22
plugins/pr-review: skip apt-get when gh is already installed

Shimada666:
## What

Skip the `apt-get update && apt-get install -y gh` step when `gh` is already on `PATH`. Behavior on runners that actually lack `gh` is unchanged.

```yaml
if command -v gh >/dev/null 2>&1; then
    echo "gh already installed: $(gh --version | head -n1)"
else
    sudo apt-get update
    sudo apt-get install -y gh
fi
```

## Why

`gh` is preinstalled on every GitHub-hosted Ubuntu runner image ([actions/runner-images](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cli-tools)), so the current unconditional install is pure overhead on hosted runners.

Measured on two back-to-back real PR reviews on `ubuntu-latest`, the `apt-get update && apt-get install -y gh` step took **~11s and ~12s** — the single largest fixed cost in the environment setup, accounting for roughly 30% of the "job start → OpenHands SDK ready" window.

As a bonus, this also stops transient `apt-get update` mirror hiccups from turning into review failures — if `gh` is already present, we never touch apt.

## Backward compat

Self-hosted runners without `gh` preinstalled see the exact same path as today (`apt-get update && apt-get install -y gh`). No input changes, no behavior change for anyone who was relying on the apt install.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Shimada666:
The failing `Tests / test_workflow_files_are_in_sync` check is **pre-existing on `main`** and unrelated to this PR.

`tests/test_workflow_sync.py` asserts that `.github/workflows/*.yml` matches the mirrored copies under `plugins/*/workflows/*.yml`. Commit [b73d2a1](https://github.com/OpenHands/extensions/commit/b73d2a1400293bcfe8b4e9ece64ed09bfc8d0d37) (#189, `chore: use OPENHANDS_BOT_GITHUB_PAT_PUBLIC instead of PAT_TOKEN`) updated the canonical files under `.github/workflows/` but did not sync the mirrored copies, so the two trees diverged on exactly one line each:

```diff
# .github/workflows/pr-review-by-openhands.yml  vs
# plugins/pr-review/workflows/pr-review-by-openhands.yml
-github-token: ${{ secrets.OPENHANDS_BOT_GITHUB_PAT_PUBLIC }}
+github-token: ${{ secrets.PAT_TOKEN }}
```

```diff
# .github/workflows/vulnerability-scan.yml  vs
# plugins/vulnerability-remediation/workflows/vulnerability-scan.yml
-github-token: ${{ secrets.OPENHANDS_BOT_GITHUB_PAT_PUBLIC || secrets.GITHUB_TOKEN }}
+github-token: ${{ secrets.PAT_TOKEN || secrets.GITHUB_TOKEN }}
```

The same job failed on the `#189` merge commit itself (see the `Tests` workflow history for `main`). This PR only touches `plugins/pr-review/action.yml`.

Happy to include the workflow-mirror resync in this PR or open a separate one — just say the word.

Shimada666:
Follow-up: opened #194 to resync the two drifted plugin workflow mirrors. Merging that will turn the `Tests` check green on `main` again — and also here, since this PR doesn't touch the affected files.

PR #240, opened 2026-05-17
skill(openhands-automation): make preset choice cost-aware

tofarr:
> _This PR was opened by an AI agent (OpenHands) on behalf of @tofarr, based on a hands-on session where the current rules steered the agent into using the LLM-backed prompt preset for a deterministic "post a quote to Slack every 5 minutes" task. The conversation that motivated this change is summarised below._

## Problem

The current `openhands-automation` skill optimises the preset choice for **functional capability** ("can a preset do this?") rather than **task fit** ("should an LLM do this?"). The four critical rules at the top of `SKILL.md` are effectively:

1. ALWAYS use a preset.
2. NEVER write custom SDK scripts unless the user explicitly asks.
3. If neither preset fits, ask the user — but the fit criteria are entirely about technical features (plugins / non-Python entrypoints / multi-file projects).
4. Same as 2.

Both presets wrap a full `Conversation` from the Agent SDK, so every scheduled run boots a sandbox, pulls the user's LLM credentials, and burns tokens. That is correct and useful for genuinely agent-shaped work (triage, summarisation, code review, on-call response).

It is **not** the right default for deterministic, high-frequency work — e.g. "post a templated message to Slack every 5 minutes" (≈ 288 LLM-driven sandbox starts/day for a job that is literally `random.choice(quotes)` + one HTTP POST). The current rules give the agent no signal to flag this trade-off, and rule 2's hard ban prevents it from even *suggesting* a script-based alternative.

## Changes

Surgical edits to **`skills/openhands-automation/SKILL.md`** only:

1. **Added a "Step 0" agent rule** at the top of the critical-rules block: *does this task need an LLM at all?* Calls out fixed data transforms, healthcheck pings, templated notifications, and rotating from a known list as patterns that should default to a custom script. Explicitly notes the cost shape (e.g. every-5-min ≈ 288 runs/day) and asks the agent to be especially careful with sub-hourly crons.
2. **Softened rule 2** from "NEVER write custom SDK scripts unless the user asks" to "do not silently create custom scripts, but *do* proactively recommend the custom path when the task is deterministic or high-frequency." Custom automations still require user consent before implementation — they just stop being invisible to the agent's decision making.
3. **Expanded the "Choosing the Right Preset" decision table** to add a row for deterministic tasks pointing at the custom-script path, and replaced the "Simple tasks with natural language prompt → Prompt Preset" row with criteria that actually describe agent-shaped work (reasoning, summarisation, triage, open-ended tool use).
4. **Named the deterministic-delivery pattern** explicitly (standup reminders, healthchecks, quote rotations, fact-of-the-day messages) and added a high-frequency tripwire: any cron tighter than hourly deserves a deliberate "should this really be agent-driven?" check.
5. **Loosened the "Reference Files" note** so the agent can consult `references/custom-automation.md` when *evaluating* the custom path. Implementing one still requires explicit user agreement.

No changes to API contracts, examples, or `references/custom-automation.md`. The intent is to recalibrate defaults, not to rewrite the skill.

## Diff stats

```
 skills/openhands-automation/SKILL.md | 36 +++++++++++++++++++++---------------
 1 file changed, 21 insertions(+), 15 deletions(-)
```

## Not in scope (worth a follow-up?)

- A first-class **`/preset/script`** endpoint (inline Python/bash on cron, no SDK / no `Conversation`) would let the skill say "for deterministic tasks, use `/preset/script`" without forcing the agent into the more error-prone custom-tarball path. That is a product change against the automation service rather than a skill change.
- Per-run cost estimation surfaced in the API response (token budget × schedule frequency) so the agent has concrete numbers to share with the user.

## Status

Drafted for review — feedback on wording, scope, and whether to fold in the follow-ups above all welcome.

---
_This PR was opened by an AI agent (OpenHands) on behalf of @tofarr._

### DIFF
--- a/skills/openhands-automation/SKILL.md
+++ b/skills/openhands-automation/SKILL.md
@@ -22,16 +22,17 @@ Create and manage automations that run in OpenHands Cloud sandboxes — triggere
 
 > **⚠️ CRITICAL — Agent behavior rules:**
 >
-> 1. **ALWAYS use preset endpoints** to create automations. They handle all SDK boilerplate, tarball packaging, and upload automatically:
->    - **Prompt preset** (`POST /v1/preset/prompt`) — for simple tasks with a natural language prompt
+> 0. **Does this task need an LLM at all? Check first.** Before picking a preset, ask whether the task actually requires reasoning, judgment, summarization, or open-ended tool use. If it is fully deterministic — fixed data transforms, scheduled HTTP calls, healthcheck pings, file rotation, picking from a known list, posting a templated message — an LLM-driven preset is overkill. Every run will spin up a sandbox and consume LLM tokens, which adds up fast at high frequencies (every 5 min ≈ 288 runs/day). Surface the trade-off to the user and offer the custom-script path (see `references/custom-automation.md`) as the cheaper, more reliable option. Be especially careful for cron schedules tighter than hourly.
+> 1. **For LLM-appropriate work, default to preset endpoints.** They handle all SDK boilerplate, tarball packaging, and upload automatically:
+>    - **Prompt preset** (`POST /v1/preset/prompt`) — for tasks expressed as a natural language prompt that benefit from agent reasoning
 >    - **Plugin preset** (`POST /v1/preset/plugin`) — when plugins with skills, MCP configs, or commands are needed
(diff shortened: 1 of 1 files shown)


malhotra5 (approved):
This is great!

PR #249, opened 2026-05-20
Add maintainer feedback collection to PR reviews

anirudhkhatry:
## Summary
- append the maintainer feedback request to the main GitHub review body instead of posting a separate PR comment
- pass `collect-feedback` and the workflow run URL into `agent_script.py` / `prompt.py` so the agent can add the inline footer itself
- collect thumbs-up/down feedback from review-body reactions during evaluation while keeping legacy issue-comment markers supported
- update the plugin README and focused tests for the new flow

## Testing
- `uv run --group test --with pyyaml pytest -q tests/test_pr_review_feedback.py tests/test_pr_review_prompt.py`

_This PR description was created by an AI agent (OpenHands) on behalf of the user._


csmith49:
Looks good, one question and one suggestion:

* Q: Since we're gathering reactions when evaluating the review, I'm assuming that means we only get reactions pre-merge?
* S: Instead of posting a whole new message, it might be better to get the agent to add some kind of foot-note to the main review to avoid spam. Which may mean feeding the content of your workflow into the initial prompt in `agent_script.py`. WDYT?

anirudhkhatry:
Re: Since we're gathering reactions when evaluating the review, I'm assuming that means we only get reactions pre-merge? Yes, that is correct. We would only get a signal when the skill is invoked. For post merge signals we can compare with the merged output? Something like reflected suggestions (based on what you did with the PR review analysis)

Re: Instead of posting a whole new message, it might be better to get the agent to add some kind of foot-note to the main review to avoid spam. Which may mean feeding the content of your workflow into the initial prompt in agent_script.py. WDYT? Yes. That would certainly be better. Let me update the PR.

PR #255, opened 2026-05-20
fix: document OH_* runtime env var fallbacks in custom-automation

tofarr:
## Problem

Custom automation scripts generated from `custom-automation.md` fail at runtime because `OPENHANDS_API_KEY` and `OPENHANDS_CLOUD_API_URL` are not present in the sandbox environment in local/dev deployments.

The dispatcher injects these in cloud deployments, but the sandbox runtime always provides equivalent values under different names:
- `OH_SESSION_API_KEYS_0` — runtime session API key
- `OH_INTERNAL_SERVER_URL` — runtime internal server URL

Notably, the workspace SDK already handles this for the session key: `OpenHandsCloudWorkspace._init_local_agent_server_mode` reads `SESSION_API_KEY` with a fallback to `OH_SESSION_API_KEYS_0`. The code examples in this skill should apply the same pattern for the constructor arguments.

## Changes

- **Both code examples*- **Both code examples*- **Both code ewith `OH_*` fallbacks instead of hard `os.environ[]` lookups
- **Env vars table** split into two sections — _Injected by the dispatcher_ and _Provided by the runtime_ — so the distinction is explicit
- **Callout note** added explaining when each set of vars is available

## Discovered via

Observing failed automation runs in a local agent-canvas dev stack, then inspecting the sandbox environment with `printenv` via the `execute_bash_command` endpoint.

_This PR was created by an AI agent (OpenHands) on behalf of the user._

### DIFF
--- a/skills/openhands-automation/references/custom-automation.md
+++ b/skills/openhands-automation/references/custom-automation.md
@@ -232,9 +232,9 @@ from openhands.sdk import Conversation
 from openhands.tools.preset.default import get_default_agent
 from openhands.workspace import OpenHandsCloudWorkspace
 
-# Environment variables are automatically injected by the dispatcher
-api_key = os.environ["OPENHANDS_API_KEY"]
-api_url = os.environ["OPENHANDS_CLOUD_API_URL"]
+# The API key and server URL each have two names depending on the deployment.
+api_key = os.environ.get("SESSION_API_KEY") or os.environ.get("OH_SESSION_API_KEYS_0", "")
+api_url = os.environ.get("RUNTIME_URL") or os.environ.get("OH_INTERNAL_SERVER_URL", "")
 
 # Use OpenHandsCloudWorkspace to connect to your OpenHands Cloud account
 with OpenHandsCloudWorkspace(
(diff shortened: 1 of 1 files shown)


skills/openhands-automation/references/custom-automation.md
```diff
@@ -290,13 +298,14 @@ The `OpenHandsCloudWorkspace` also has a `keep_alive` parameter, but in `local_a
...
-| Variable | Description |
-|----------|-------------|
-| `OPENHANDS_API_KEY` | API key for OpenHands services |
-| `OPENHANDS_CLOUD_API_URL` | Base URL for the OpenHands Cloud API |
-| `AUTOMATION_EVENT_PAYLOAD` | JSON with trigger info, automation ID, and name |
-| `SANDBOX_ID` | The sandbox ID where the automation is running |
-| `SESSION_API_KEY` | Session API key for sandbox operations |
+| Variable | Aliases | Description |
+|----------|---------|-------------|
+| `OPENHANDS_API_KEY` | `SESSION_API_KEY`, `OH_SESSION_API_KEYS_0` | API / session key — all three are the same credential, different deployments set different names |
```
enyst: This is interesting! Is this… really the same, OPENHANDS_API_KEY is used in workflows and automation as the cloud app-server key (the credential the user can create from API keys page), while the other two are agent-server 🤔

tofarr: I think you're right that these are not the same. This code is run inside the same container as the agent server.

In local: we have `OH_SESSION_API_KEYS_0` and `OH_INTERNAL_SERVER_URL`
In cloud, we have `SESSION_API_KEY` and `RUNTIME_URL`

Neither of these define `OPENHANDS_API_KEY` or `OPENHANDS_CLOUD_API_URL`

I think we can remove these references.

tofarr: Good catch — you are right. `OPENHANDS_API_KEY` / `OPENHANDS_CLOUD_API_URL` are not injected in either environment. Updated to use only the real vars:

- Local: `OH_SESSION_API_KEYS_0` / `OH_INTERNAL_SERVER_URL`
- Cloud: `SESSION_API_KEY` / `RUNTIME_URL`

All references to `OPENHANDS_API_KEY` and `OPENHANDS_CLOUD_API_URL` have been removed.

PR #317, opened 2026-06-08
fix: replace colons in command filenames for Windows compatibility

### LINKED ISSUE #230: Colon `:` in file names breaks git sync on Windows
Ran into a small issue introduced in #174. A file name now contain a colon `:`: `skills/openhands-automation/commands/automation:create.md`, which is an invalid character on Windows. This makes it impossible to sync the repo locally:

```
sergey@pluto MINGW64 /w/code/OpenHands/extensions (main)
$ gh repo sync
failed to run git: error: invalid path 'skills/openhands-automation/commands/automation:create.md'
```

Would it be possible to rename those files using a different separator? I get that the tooling itself runs on Unix-like envs, but it'd be nice to be able to clone and browse the sources on Windows without issues.

Thanks!

jamiechicago312:
Fixes #230

### Problem
Colons in filenames (e.g., `automation:create.md`) break git sync on Windows.

### Solution
1. Replace `:` with `-` in `scripts/sync_extensions.py`
2. Rename 8 existing colon-named files

---
*Created by AI agent on behalf of @jamiechicago312*


jamiechicago312 (commented):
Nice fix. Replacing `:` with `-` in generated command filenames keeps the change small and easy to maintain, while the command trigger itself still stays unchanged. I pushed one follow-up commit (`84a2f6c`) that adds a focused regression test so this Windows-compatibility behavior stays covered going forward.

Validation I ran:
- `.venv/bin/python -m pytest -q tests/test_sync_extensions.py`
- `.venv/bin/python scripts/sync_extensions.py --check`

_This review was created by an AI agent (OpenHands) on behalf of the user._

jpelletier1:
@malhotra5 I think there are other examples where colons are used in SKILLS related to pre-built automations, such as `github-repo-monitor`. I think we need to make sure this is fixed across the pre-built extensions.

malhotra5 (approved):
LGTM

malhotra5:
Also just to note, in the latest versions of agent canvas we no longer git clone the extensions repo. Rather we bundle the npm package for the extensions repo directly into the agent canvas. Nevertheless this seems like a good fix to get in

PR #322, opened 2026-06-09
docs: harden slack-channel-monitor setup instructions

### LINKED ISSUE #321: Harden slack-channel-monitor setup instructions against template reimplementation
## Summary
The `skills/slack-channel-monitor` skill works, but its setup instructions could be more explicit for weaker models. In a recent setup flow, the model produced a hand-written simplified polling script instead of copying `scripts/main.py` and changing only the documented constants. That caused multiple avoidable failures even though the shipped template already had the correct logic.

## Why this matters
The current skill is functionally correct, but it leaves enough room for a weaker model to:
- reimplement the script instead of copying the template
- drop required helpers like `get_secret()`
- change state-path logic
- send the wrong conversation creation payload

Users should not need a top-tier model just to configure this skill reliably.

## Proposed hardening
Update the skill instructions to:
1. explicitly require copying `scripts/main.py` verbatim
2. explicitly forbid reimplementing or simplifying the script
3. add a small integrity checklist after customization to confirm key helpers are still present

This would keep runtime behavior unchanged while making the setup path more robust across models.

_AI-generated issue created by OpenHands on behalf of the user._

jamiechicago312:
# Human Generated

When i used minimax m2.7 it failed to create the automation. gpt 5.4 had to go back and fix it.
This PR should help harden the slack automation skill so that smaller models will be successful on the first try

---
## Summary
- explicitly require copying `skills/slack-channel-monitor/scripts/main.py` verbatim
- forbid reimplementing or simplifying the polling script during setup
- add a quick integrity checklist so weaker models verify the key helper functions are still present

## Testing
- `python3 scripts/sync_extensions.py --check`

Closes #321

_AI-generated pull request created by OpenHands on behalf of the user._


jamiechicago312:
Updated the PR title to conventional-commit format so the pr-title check passes.\n\n_This comment was generated by an AI agent (OpenHands) on behalf of the user._

enyst (approved):
We could try, thank you!

PR #395, opened 2026-07-08
fix: make root marketplace catalog real files (not symlinks)

### LINKED ISSUE #394: Marketplace catalog unreachable via symlinked .claude-plugin path (breaks plugin-directory / `github://openhands/extensions`)
## Summary

When the [plugin-directory](https://github.com/openhands/plugin-directory) is configured with `MARKETPLACE_SOURCE=github://openhands/extensions`, the marketplace catalog **does not render** — the Plugin Directory shows an empty catalog.

Root cause: the catalog is exposed through a **double symlink**, and `raw.githubusercontent.com` does **not** dereference symlinks through directory paths.

## How plugin-directory fetches the catalog

`openhands/plugin-directory` builds a hardcoded raw URL from the source config (`server/app/services/marketplace_service.py`):

```python
CATALOG_PATH = ".claude-plugin/marketplace.json"
url = f"https://raw.githubusercontent.com/{owner}/{repo}/{ref}/{catalog_path}"
```

So it requests exactly:

```
https://raw.githubusercontent.com/openhands/extensions/main/.claude-plugin/marketplace.json
```

## The symlink chain in this repo

Traced via the GitHub git Trees API (`mode=120000` = symlink):

| Path | Git mode | Type | Target |
|------|----------|------|--------|
| `.claude-plugin` | `120000` | symlink | `.plugin` |
| `.plugin/marketplace.json` | `120000` | symlink | `../marketplaces/openhands-extensions.json` |
| `marketplaces/openhands-extensions.json` | `100644` | real file | (the actual catalog) |

## Verification

`raw.githubusercontent.com` serves the raw blob content stored in git. A symlink is stored as a blob whose content is the link-target text, and the raw server does **not** traverse symlinked directory paths:

- `.claude-plugin/marketplace.json` → **HTTP 404** (symlinked directory not traversable)
- `.plugin/marketplace.json` → returns the **literal string** `../marketplaces/openhands-extensions.json` (the symlink blob body), not the JSON catalog — so even if level 1 resolved, JSON parsing would fail
- `marketplaces/openhands-extensions.json` → **HTTP 200**, the real catalog JSON

As a result, plugin-directory's `fetch_catalog()` raises `CatalogUnavailableError` on the 404, the plugins list comes back empty, and nothing renders. This is a **server-side (GitHub raw) limitation**, not something the plugin-directory client can work around — the catalog simply never arrives.

## Suggested fix

1. **In this repo (simplest):** commit a real `.claude-plugin/marketplace.json` file (or a real `.claude-plugin/` directory) instead of the symlink chain, so the catalog is reachable at the path plugin-directory (and other consumers using the standard `.claude-plugin/marketplace.json` convention) expects.

Happy to open a PR for either direction if helpful.

---
_This issue was created by an AI agent (OpenHands) on behalf of a user._

jpelletier1:
- [ ] A human has tested these changes.

---

## Why

When the [plugin-directory](https://github.com/openhands/plugin-directory) is configured with `MARKETPLACE_SOURCE=github://openhands/extensions`, the marketplace catalog **does not render** (empty Plugin Directory).

Root cause: the catalog was exposed through a **double symlink** chain, and `raw.githubusercontent.com` does **not** dereference git symlinks (it serves the literal symlink target text; for symlinked directory paths it returns 404):

- `.claude-plugin` (symlink) -> `.plugin`
- `.plugin/marketplace.json` (symlink) -> `../marketplaces/openhands-extensions.json`

Verified empirically against `main`: `GET .plugin/marketplace.json` returns HTTP 200 but the body is the literal string `../marketplaces/openhands-extensions.json` (not JSON), and `GET .claude-plugin/marketplace.json` returns 404 (symlinked directory not traversable). So consumers doing `GET` + `json.loads()` get a 404 or a parse error and an empty catalog. This is a server-side (GitHub raw) limitation the plugin-directory client cannot work around, so the catalog simply never arrives.

## Summary

- Replaced the root vendor symlinks (`.claude-plugin`, `.codex-plugin`, `.plugin/marketplace.json`) with **real directories + real `marketplace.json` files** that each mirror the canonical `marketplaces/openhands-extensions.json` byte-for-byte (canonicalized JSON). Git mode changed from `120000` (symlink) to `100644` (regular file). The catalog is now reachable at every standard `.claude-plugin/marketplace.json` / `.codex-plugin/marketplace.json` / `.plugin/marketplace.json` path over the raw URL.
- Added a `root-marketplace` sync task to `scripts/sync_extensions.py` that keeps all three root copies in sync with the canonical file and reports drift in `--check` mode. Since CI runs `python scripts/sync_extensions.py --check` with no task filter, the new task is automatically enforced on every PR.
- Added tests in `tests/test_sync_extensions.py` (`TestRootMarketplace`) verifying the real-repo files are non-symlinks, match the canonical catalog, are valid JSON, and that the sync task detects/fixes both the original symlink chain and stale copies.
- Updated `AGENTS.md` with the invariant (root catalog copies are real files, never symlinks).

## Issue Number

Closes #394

## How to Test

1. `pip install pyyaml pytest`
2. `python scripts/sync_extensions.py --check` - exits 0 (only the pre-existing, non-blocking coverage warning for `plugins/issue-duplicate-checker` remains).
3. `python -m pytest tests/test_sync_extensions.py -q` - 40 passed.
4. Confirm the files are real (mode `100644`, not `120000`): `git ls-files -s -- .claude-plugin .codex-plugin .plugin`
5. (Optional, network) After merge to `main`, `curl -fsSL https://raw.githubusercontent.com/OpenHands/extensions/main/.claude-plugin/marketplace.json` returns the catalog JSON (HTTP 200) instead of 404. _(Cannot verify pre-merge from this branch since the raw URL serves `main`.)_

## Video/Screenshots

N/A - file-system / raw-URL fix with no UI surface.

## Notes

- The `.claude-plugin/marketplace.json`, `.codex-plugin/marketplace.json`, and `.plugin/marketplace.json` files are now generated copies; edit the canonical `marketplaces/openhands-extensions.json` and run `python scripts/sync_extensions.py` (no args) to refresh all three. Do not reintroduce symlinks at these paths.
- Per-skill/per-plugin `.claude-plugin` and `.codex-plugin` vendor symlinks *within* `skills/` and `plugins/` are unrelated and unchanged (managed by the existing `symlinks` sync task).
- JSON is canonicalized (sorted keys, 2-space indent) for stable, minimal diffs and reliable byte-for-byte equality checks.

_This PR was created by an AI agent (OpenHands) on behalf of a user._


jpelletier1:
@jpshackelford As someone passionate about Plugins, I wanted to see if you had a chance to review this PR. Specifically, the symlinks we use in the OpenHands/extensions repo makes it difficult to render the full plugin marketplace in the Plugin Directory -- this attempts to fix that by creating an actual marketplace.json in the respective folders and keeping them in sync.

PR #423, opened 2026-07-27
feat(automations): add the setup contract to the automation catalog

### LINKED ISSUE #420: Define reference automation manifests and API contract fixtures
## Context

OpenHands/extensions is the source of truth for automation-specific UI, setup, and funnel semantics. Agent Canvas and the automation service need concrete examples before implementing their generic contracts.

These fixtures are contract inputs, not a separate runtime or installable application.

## Acceptance criteria

* Add extension-owned examples for direct scheduled setup, direct GitHub-event setup, and assisted conversation setup.
* Each example defines its UI, validation/error states, preflight and creation requests, and stable analytics stages/properties without secrets.

hieptl:

---

## Why

Three repos need to build the extension-owned Automation UI in parallel and none of them can start, because the contract between them does not exist.

- **agent-canvas** must load and validate an automation's setup definition, but there is no schema for one.
- **automation** must expose capabilities discovery and preflight validation, but there is no agreed request or response shape. Neither endpoint exists today - a grep for `capabilit|preflight|dry.?run|/validate` across `openhands/automation/` returns only a docstring and a CORS comment.
- **extensions** owns all per-automation UI, copy, fields, request mappings, and analytics, but ships only `automations/catalog/*.json` - nine fields with no UI information, whose `prompt` is a slash command handed to an LLM to interpret.

This also needed verification rather than drafting. The create payload sketched in the project reference document is **rejected by the live API**:

```
repos.0 -> Value error, Short URL format 'OpenHands/agent-canvas'
           requires explicit 'provider' field.
```

`RepoSource` requires `provider` for short `owner/repo` URLs and `CreatePromptAutomationRequest` is declared `extra="forbid"`, so this is a hard 422, not a silently dropped field. Canvas's own import path already works around it independently at `automation-service.api.ts:153-154`. Two teams were about to implement the documented version and discover the mismatch at the first user's first click.

## Summary

- Add a nested `setup` block to `automations/catalog/*.json` - the extension-owned configuration experience for one automation - and `automations/catalog.schema.json` to validate the whole entry. The schema is the trust boundary: closed `submit.action` enum, service-relative `^/v1/` paths, in-app redirects only, no markup in copy, namespaced placeholders, and `requires.secrets` entries with no field able to hold a value. `setup` is optional and cannot restate `id`, `name`, `category`, or `description`, so one automation stays one hand-authored record.
- Fill in three entries covering the setup archetypes - `github-pr-reviewer` (direct scheduled), `github-repo-monitor` (direct GitHub-event with a JMESPath filter), `incident-retrospective-drafter` (assisted conversation) - plus contract fixtures pairing form values with the exact request bodies they must produce, across 10 scenarios and 3 deployment capability shapes.
- Package it the way integrations are packaged: `scripts/build-automation-catalog.mjs` generates `automations/catalog-index.js`, and `automations/index.js` becomes a stable API (`AUTOMATION_CATALOG`, `listAutomationCatalog()`, `getAutomationCatalogEntry(id)`). Fixtures are test vectors, so they live in `tests/fixtures/automations/`, are optional per automation, and are reachable only through the explicit `@openhands/extensions/testing/automations/*.json` subpath - never from the runtime entry point.

## Issue Number

Resolves https://github.com/OpenHands/extensions/issues/420

## How to Test

```bash
# 1. Schema, entries, fixtures, and the form-values -> request mapping
uv sync --group test
uv run pytest -q tests/test_automation_setup.py        # 44 passed, 1 skipped

# 2. No regressions in the existing suite
uv run pytest -q tests/ --deselect tests/test_live_integration_smoke.py
#    508 passed, 1 skipped

# 3. The generated index is reproducible and current
npm run build:automations && git diff --exit-code automations/catalog-index.js

# 4. The stable API resolves and setup rides along on the entry
node --input-type=module -e "
  import { AUTOMATION_CATALOG, listAutomationCatalog, getAutomationCatalogEntry }
    from './automations/index.js';
  console.log(AUTOMATION_CATALOG.length, listAutomationCatalog().length);
  console.log(getAutomationCatalogEntry('github-pr-reviewer').setup.mode);
  console.log(listAutomationCatalog().filter(e => e.setup).map(e => e.id));
"

# 5. Repo sync check still passes
python scripts/sync_extensions.py --check
```

**Verification against the live service.** Every fixture request body was validated against the real Pydantic models in a local checkout of `OpenHands/automation`:

- each `201` body validates against `CreatePromptAutomationRequest`
- each `422` body reproduces the error detail the service actually returns, including the `loc` path - note it carries the discriminated-union tag, `["body","trigger","cron","schedule"]`, not `["body","trigger","schedule"]`
- each response `trigger` matches what `model_dump()` stores, so materialised defaults such as `timezone` are present
- each event filter compiles under `validate_filter()`
- every preflight `draft` is itself a well-formed create body

**Package resolution.** Linking the working tree into a scratch `node_modules` confirms the new subpaths resolve and the old ones no longer do:

```
fixtures: github-pr-reviewer 4 scenarios
capabilities shapes: supported, cronOnly, notReady
schema title: AutomationCatalogEntry
api: 8 8 direct
removed subpaths no longer resolve
```

I also confirmed the tests fail on real breakage rather than passing vacuously: dropping `provider` from an entry's payload, or removing one `errorMap` entry, each turns a test red.

## Video/Screenshots

None. This PR adds no UI - it ships JSON data files, a schema, a codegen script, and tests. Nothing loads a `setup` block yet; verification is the command output above.


neubig (commented):
Thanks, Hiep — overall this is thoughtful work, especially the schema constraints and API verification.

I think the automation packaging would be simpler and less prone to drift if it followed the existing integrations design more closely:

- Keep one hand-authored source file per automation under `automations/catalog/*.json`.
- Put both the card metadata and a nested `setup` definition in that file; `setup` would not repeat `id`, `name`, `category`, or `description`.
- Generate a static `catalog-index.js` from the directory, following `scripts/build-integration-catalog.mjs`.
- Keep `automations/index.js` as a stable API exposing `AUTOMATION_CATALOG`, `listAutomationCatalog()`, and `getAutomationCatalogEntry(id)`.
- Keep fixtures as independent test vectors, but in the testing directory, and do not require them for every automation. Do not import them from the normal runtime entry point. If downstream contract tests need package access, they could live behind an explicit testing subpath.

Concretely, the production layout would be approximately:

```
automations/
  catalog/*.json          # single source of truth
  catalog.schema.json
  catalog-index.js        # generated
  index.js
  index.d.ts
```

This would make adding or changing an automation a one-JSON-file operation and avoid catalog/manifest inconsistencies. The documented cron-versus-webhook difference for `github-repo-monitor` is a good example of the drift this structure would prevent.

Type generation from the schema would also be valuable, but that can reasonably remain a follow-up.

hieptl:
Hello @neubig,

I've addressed your feedback and updated the code accordingly.

When you have a chance, could you please review the pull request again?

Thank you very much! 🙏

automations/catalog/github-pr-reviewer.json
```diff
@@ -9,5 +9,269 @@
...
+  "exampleImplementation": "Trigger: cron polling for open GitHub PRs with a configured label such as openhands-review\nRequired secret: GITHUB_PERSONAL_ACCESS_TOKEN\n\n1. Read repository, trigger label, review tone, and polling schedule from setup.\n2. List open PRs and find the latest matching GitHub labeled issue event for each labeled PR.\n3. Deduplicate on the label event ID so every label application queues exactly one review.\n4. Start an OpenHands conversation that clones the repo, checks out the exact PR head SHA, and inspects PR discussion, review comments, changed files, diff, and surrounding code.\n5. Post an acknowledgement with the conversation link, then post the final AI review comment only if the PR is still open and the head SHA has not changed.",
+  "setup": {
+    "version": "1.0",
+    "mode": "direct",
+    "routes": [
+      {
+        "path": "/automations/new/github-pr-reviewer",
+        "page": "setup"
+      }
+    ],
```
neubig: Could we delete this element and just rely on the name of the automation?

automations/catalog/github-pr-reviewer.json
```diff
@@ -9,5 +9,269 @@
   "popularityRank": 100,
   "estimatedSetupMinutes": 4,
   "prompt": "/pr-reviewer:setup",
```
neubig: Let's match this with the name

```suggestion
  "prompt": "/github-pr-reviewer:setup",
```

automations/catalog/github-pr-reviewer.json
```diff
@@ -9,5 +9,269 @@
...
+  "setup": {
+    "version": "1.0",
+    "mode": "direct",
+    "routes": [
+      {
+        "path": "/automations/new/github-pr-reviewer",
+        "page": "setup"
+      }
+    ],
+    "capabilities": {
```
neubig: I think "capabilities" can be deleted.

- "discovery": is always the same, so it is unnecessary
- "requires": there is also a "requires" dictionary one level up in the hierarchy, so we could move "triggerKinds" and "features" there.
- "onUnsupported": can be deleted and just generated programmatically.

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -9,5 +9,269 @@
...
+          "constraint": "options",
+          "from": "triggers.cron.timezones"
+        }
+      ],
+      "onUnsupported": {
+        "behavior": "block",
+        "message": "This deployment cannot run scheduled automations, so the code review agent is unavailable here."
+      }
+    },
+    "requires": {
```
neubig: "reason", " help", " and "message" all mean the same thing but with different terms. Let's unify on "message".

automations/catalog/github-pr-reviewer.json
```diff
@@ -9,5 +9,269 @@
...
+        "behavior": "block",
+        "message": "This deployment cannot run scheduled automations, so the code review agent is unavailable here."
+      }
+    },
+    "requires": {
+      "integrations": [
+        {
+          "id": "github",
+          "reason": "Used to read pull requests and post review comments.",
+          "enforcement": "block"
```
neubig: If `"enforcement": "block"`, `"behavior": "block"`, and `"required": true` mean the same thing, let's unify on `"required": true`, unless there are more than 2 possibilities for behavior, in which case we can use `"enforcement": "block"`

automations/catalog/github-pr-reviewer.json
```diff
@@ -9,5 +9,269 @@
...
+        }
+      ],
+      "secrets": [
+        {
+          "key": "GITHUB_PERSONAL_ACCESS_TOKEN",
+          "label": "GitHub personal access token",
+          "help": "Needs read access to the repository and permission to comment on pull requests.",
+          "required": true
+        }
+      ],
```
neubig: This should not be necessary if the github integration is present.

automations/catalog/github-pr-reviewer.json
```diff
@@ -9,5 +9,269 @@
...
+      "onUnmet": {
+        "behavior": "block",
+        "message": "Connect GitHub and provide an access token before continuing."
+      }
+    },
```
neubig:
```suggestion
```

Let's generate this programmatically instead of putting it in the file.

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -9,5 +9,269 @@
...
+          "help": "Needs read access to the repository and permission to comment on pull requests.",
+          "required": true
+        }
+      ],
+      "onUnmet": {
+        "behavior": "block",
+        "message": "Connect GitHub and provide an access token before continuing."
+      }
+    },
+    "form": {
```
neubig, line 23: I would like to separate two concepts:

1. triggers: how the `OpenHands/automations` script decides whether to run the automation or not. in this case it is "cron", with all of the settings that a cron job needs, so there would be a dictionary with "cron": and then a list of all the arguments needed for a cron-based trigger.

2. args: all of the arguments to the automation script/prompt, which will be everything else in "fields"

Upon writing this, I realized that we will also need to include an automation script in this catalog, since automations also upload tarballs. So maybe each automation in the catalog can be a directory, with "manifest.json" and then any of the other scripts that get uploaded as a ".tar.gz" to the automations service.

neubig (commented):
Thanks @hieptl , I left some comments on the automations catalog format, please take a look! Apologies for leaving a partial review and not reviewing the whole thing, because I need to run out of the door now, but hopefully this gives you some stuff to work on, and once this is done I think we'll be pretty close to finished for this PR.

hieptl:
Hello @neubig,

I've addressed your feedback and updated the code accordingly.

When you have a chance, could you please review the pull request again?

Thank you very much! 🙏

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
...
+          },
+          {
+            "name": "timezone",
+            "type": "timezone",
+            "label": "Timezone",
+            "help": "Timezone the schedule is interpreted in.",
+            "default": "UTC",
+            "required": true
+          }
+        ]
```
neubig:
```suggestion
        "cron": {
          "schedule" : {
            "type": "cron",
            "label": "Check frequency",
            "help": "How often to look for newly labelled pull requests.",
            "default": "*/15 * * * *",
            "required": true
          },
          "timezone" : {
            "type": "timezone",
            "label": "Timezone",
            "help": "Timezone the schedule is interpreted in.",
            "default": "UTC",
            "required": true
          }
        }
```

A dict may be better here.

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
...
+      },
+      "args": [
```
neubig: Ditto, a dict may be better here.

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
...
+          "trigger.schedule": "schedule",
+          "trigger.timezone": "timezone",
+          "prompt": [
+            "triggerLabel",
+            "reviewTone",
+            "repository"
+          ]
+        }
+      }
+    },
```
neubig: Is this necessary? I feel like we could automatically decide the behavior from the configuration of the arguments, etc.

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
...
+  "id": "github-pr-reviewer",
+  "name": "GitHub Code Review Agent",
+  "category": "Code review",
+  "description": "Watch for a configurable label on GitHub pull requests, inspect full PR and repository context, and post an AI review comment once per label event.",
+  "requiredIntegrationIds": [
+    "github"
+  ],
+  "popularityRank": 100,
+  "estimatedSetupMinutes": 4,
+  "prompt": "/github-pr-reviewer:setup",
```
neubig: Maybe this can be removed and automatically derived from the automation id?

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
+{
+  "id": "github-pr-reviewer",
+  "name": "GitHub Code Review Agent",
+  "category": "Code review",
+  "description": "Watch for a configurable label on GitHub pull requests, inspect full PR and repository context, and post an AI review comment once per label event.",
+  "requiredIntegrationIds": [
+    "github"
+  ],
```
neubig: This is duplicated, since the information already exists elsewhere.

```suggestion
```

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
...
+          "value": "{{form.reviewTone}}"
+        },
+        {
+          "label": "Runs",
+          "value": "{{form.schedule}} ({{form.timezone}})"
+        }
+      ],
+      "confirmLabel": "Create automation"
+    },
+    "submit": {
```
neubig: I think we can remove this and derive the behavior from other fields we already have.

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
...
+          "trigger.timezone": "timezone",
+          "prompt": [
+            "triggerLabel",
+            "reviewTone",
+            "repository"
+          ]
+        }
+      }
+    },
+    "review": {
```
neubig: I'm not sure what this field stands for, but maybe it can be removed and derived from fields we already have?

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,235 @@
...
+        "behavior": "navigate",
+        "to": "/automations/{{response.id}}"
+      },
+      "onError": {
+        "behavior": "stayOnForm",
+        "errorTarget": "field",
+        "reuseErrorMap": true
+      }
+    },
+    "analytics": {
```
neubig: I think this field can be removed and just implemented as standardized code that is shared between all automations.

neubig (commented):
I left some more comments.

As a general principle, I would like the manifests to have the minimal possible amount of information needed to uniquely specify the automation. Anything that varies from automation to automation needs to be specified there, but *only once*. Any fields that can be derived from other fields can be deleted and generated programmatically.

automations/catalog/github-pr-reviewer/manifest.json
hieptl: The implementation has been reduced to only the parts that vary. The `action`, `endpoint`, `onSuccess`, and `onError` properties were identical across all automations, so they have been removed.

What remains is the request body itself. For direct entries, this is now defined as `setup.payload`, while assisted entries use `setup.message`. I intentionally stopped short of abstracting this further because it represents the mapping from form values to the API request body, and I couldn't identify a reliable way to derive it automatically.

That said, if your vision was for the request body to be generated directly from the form as well, I'd be interested to hear how you envision that working.

Regardless of the approach, the schema continues to enforce the correct pairing: direct entries must define `payload` and cannot define `message`, while assisted entries must define `message` and cannot define `payload`.

Thank you very much! 🙏

automations/catalog/github-pr-reviewer/manifest.json
hieptl: Sorry, I didn't explain that clearly. I was referring to the confirmation screen displayed before anything is created, which consisted of a title, one row for each user-configured item, and a confirm button. I've removed it because those rows can now be derived directly from the declared fields and their labels.

One detail worth calling out is that the previous hand-crafted summaries grouped related fields into a single line—for example, `Repository: OpenHands/agent-server-gui (main)` or `Runs: */15 * * * * (UTC)`. With the derived version, each field is displayed on its own row instead. While this is slightly less concise, I considered it a presentation concern for the host application rather than something each entry should be responsible for defining.

I've captured this as an open question in the README. If you would prefer to retain a mechanism for expressing grouped summaries, I'd be happy to explore what that could look like.

Thank you very much! 🙏

automations/catalog/github-pr-reviewer/manifest.json
hieptl: Removed, thank you—that was indeed duplicated information.

I did want to validate the derivation approach with you, though, because I found that deriving it from the `id` alone doesn't quite work in the current setup.

There are two reasons for this. First, three of the eight automation IDs don't match the corresponding skill directories:

* `incident-retrospective-drafter` → `skills/incident-retrospective`
* `linear-triage-assistant` → `skills/linear-triage`
* `research-brief-writer` → `skills/research-brief`

Second, `skills/jira-issue-to-pr` doesn't define a slash trigger at all—it is invoked via its description instead. As a result, `/<id>:setup` would only resolve correctly for four of the eight automations unless the published triggers in `skills/` were renamed.

Instead, I updated the implementation so that each entry references the **skill**, and the command is derived from that skill's own `triggers:` frontmatter, which is already exposed by the skills catalog.

```json
"skill": "incident-retrospective"
```

The `skill` property defaults to `id`, so only those three exceptions need to specify it explicitly; the other five require no additional configuration. This also means the command string now has a single source of truth in the repository, allowing a skill to rename its trigger without leaving a stale copy behind here.

Does this align with what you had in mind, or would you prefer to standardize the skill triggers as `/<id>:setup` so that the `id` alone is sufficient? I kept this PR scoped to avoid changes under `skills/`, since renaming published commands felt like a separate change, but I'm happy to take either approach if you think that would be the better direction.

Thank you very much! 🙏

hieptl:
Hello @neubig,

I've addressed your feedback and updated the code accordingly.

When you have a chance, could you please review the pull request again?

Thank you very much! 🙏

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,108 @@
...
+          "type": "text",
+          "label": "Base branch",
+          "help": "Branch checked out when the review runs.",
+          "default": "main",
+          "required": true,
+          "constraints": {
+            "minLength": 1,
+            "maxLength": 255
+          }
+        },
```
neubig: I think maybe the review will run on the PR branch, so do we need this argument? Maybe it can be deleted


```suggestion
```

automations/catalog/github-repo-monitor/manifest.json
```diff
@@ -0,0 +1,97 @@
...
+              "format": "safeExpressionLiteral"
+            }
+          }
+        }
+      },
+      "args": {
+        "repository": {
+          "type": "repo-picker",
+          "label": "Repository",
+          "help": "The repository to watch for mentions.",
```
neubig: We might want to support multiple repositories

```suggestion
          "help": "The repository or repositories to watch for mentions.",
```

automations/catalog/github-pr-reviewer/manifest.json
```diff
@@ -0,0 +1,108 @@
...
+          "provider": "github"
+        }
+      ],
+      "trigger": {
+        "type": "cron",
+        "schedule": "{{form.schedule}}",
+        "timezone": "{{form.timezone}}"
+      }
+    }
+  }
```
neubig: Much of this seems duplicated with the other entries in the manifest. Either delete, or keep only the necessary entries (e.g. "example-prompt") as a top-level item.

automations/catalog/jira-issue-to-pr/manifest.json
```diff
@@ -3,12 +3,13 @@
...
   "category": "Project management",
   "description": "Watch a Jira Cloud project for issues with a configurable label and automatically open a GitHub pull request for each new issue found. The target GitHub repo is read from the ticket body - no repo parameter required at deploy time.",
-  "requiredIntegrationIds": [
-    "jira",
-    "github"
-  ],
+  "requires": {
+    "integrations": {
+      "jira": {},
+      "github": {}
```
neubig: These are empty, should they be rejected by the schema?

neubig (commented):
Looking very good, just a few final comments and I think this should be ready to approve!

automations/catalog/jira-issue-to-pr/manifest.json
hieptl: They should, and they do now. The `message` field is now required for every integration, so it's no longer possible to define an integration with an empty object. I've also added a brief explanation for each integration in the five card-only entries to clarify its purpose.

The `required` field remains optional because it defaults to `true`, and only the Notion integration used by the retrospective drafter needs to specify `false` explicitly. That said, I'm happy to make `required` explicit for every integration if you think that makes the configuration clearer or more consistent.

Thank you very much! 🙏

automations/catalog/github-pr-reviewer/manifest.json
hieptl: You're right—most of it was simply restating information that already existed elsewhere in the manifest. As a result, `payload` has been removed.

Here's where each piece of information now comes from:

* `name` is derived from the entry's `name` together with the selected repository. For example, the card now generates `GitHub Code Review Agent - OpenHands/agent-server-gui` instead of relying on a hand-written value such as `PR Reviewer - ...`.
* `repos` is derived from the repository picker field and its declared `provider`.
* `trigger` is reconstructed from the key under `form.triggers` along with the fields defined within that trigger.

The remaining declared fields are `prompt` and, for event triggers, `filter`. I left `filter` in place because it's the one part of an event trigger that can't be derived directly from the form—it composes form values into a JMESPath expression.

To make this reconstruction based on explicit mappings rather than naming conventions, I also renamed two fields in `github-repo-monitor` to match the trigger properties they populate: `eventType` is now `on`, and `baseRef` is now `ref`. I left `triggerPhrase` unchanged, since it serves as an input to `filter` rather than representing a trigger property itself.

As a result, `github-pr-reviewer/manifest.json` is now 82 lines, down from 220 when you first reviewed it.

To help ensure the reconstruction is correct, `tests/test_automation_setup.py` now rebuilds the complete create request body and verifies it against the recorded fixtures. I also revalidated each fixture against the current `CreatePromptAutomationRequest`.

Does keeping `filter` as an explicit declaration align with what you had in mind, or were you envisioning a way to derive that as well?

Thank you very much! 🙏

hieptl:
Hello @neubig,

I've addressed your feedback and updated the code accordingly.

When you have a chance, could you please review the pull request again?

Thank you very much! 🙏

neubig (approved):
Looks great now, thank you!

PR #425, opened 2026-07-28
feat(skills): carry a category on every catalog skill

sleeyax:

---

## Why

The OpenHands Skills page (`/skills`) shows every skill in one flat grid. With 58 public skills in
this repo alone, search is the only way to narrow it down, which only helps when you already know
what you're looking for. See OpenHands/OpenHands#16123 for the full motivation.

To group skills by topic, the UI needs a topical axis, and the catalog is the only place that can
own it. Marketplace entries already carry a `category`, but the values were ad-hoc and inconsistent
across manifests (`development`, `productivity`, `integration`, `monitoring`, `documentation`, …),
and they were never exported from the generated catalog, so the frontend could not read them.

## Summary

- Normalise `category` on every skill entry across `marketplaces/*.json` onto a fixed nine-value
  taxonomy: `automations`, `environment`, `code-hosting`, `agent-authoring`, `code-quality`,
  `integrations`, `writing`, `design`, `other`.
- `scripts/build-skills-catalog.mjs` now reads those manifests, validates each skill's category
  against the taxonomy, fails the build on an unknown or conflicting value, and emits `category`
  on every `SKILLS_CATALOG` entry plus a `SKILL_CATEGORY_IDS` export. Skills with no marketplace
  entry fall back to `other` and are listed in a build warning.
- Type the new field and export `SkillCategoryId` / `SKILL_CATEGORY_IDS` from `skills/index.d.ts`.
- Document the category requirement in the "Adding a Skill" section of the README.

## Issue Number

OpenHands/OpenHands#16123

## How to Test

```bash
node scripts/build-skills-catalog.mjs   # regenerates skills/index.js; should be a no-op on this branch
python3 -m pytest tests/test_skills_catalog.py -q
```

Verify the generated catalog carries the field:

```bash
node -e "import('./skills/index.js').then(m => console.log(m.SKILLS_CATALOG.slice(0,3).map(s => [s.name, s.category])))"
```

Negative paths covered by the test suite: an unknown category value and the same skill given
conflicting categories in two manifests both throw at build time.

Result of `python3 -m pytest tests/test_skills_catalog.py -q`: **38 passed**.

## Video/Screenshots

n/a — build-time/catalog change with no UI of its own. The consuming UI is shown in
OpenHands/OpenHands#16124.

## Notes

- **`category` on skill entries now means something different from `category` on plugin entries.**
  Only entries whose `source` starts with `./skills/` are re-keyed onto this taxonomy; plugin
  entries keep their existing free-form values for Claude Code marketplace browsing. If anything
  downstream groups *skill* entries by the old values, this changes what it sees. Flagging for
  reviewers who know the marketplace consumers better than I do — happy to move this onto a
  separate `skill_category` field instead if the shared key is a problem.
- Two skills (`qa-changes`, `release-notes`) have no marketplace entry and so land in `other`.
  Adding entries would mean creating `.plugin/plugin.json` plus vendor symlinks, i.e. publishing
  them as Codex/Claude Code plugins — out of scope here, and the fallback is deliberate rather
  than accidental.
- `buildCatalog()` takes an optional `marketplacesDir` so fixture-based tests can point at an empty
  directory instead of inheriting categories from this repo's real manifests.
- The consumer is OpenHands/OpenHands#16124; it degrades to `other` for any category it doesn't
  know, so a newer catalog against an older frontend renders rather than breaking.


tests/test_skills_catalog.py
```diff
@@ -334,3 +349,198 @@ def test_index_is_up_to_date(self):
...
+    "environment": 10,
+    "automations": 9,
+    "code-hosting": 8,
+    "agent-authoring": 8,
+    "code-quality": 6,
+    "integrations": 5,
+    "writing": 4,
+    "design": 2,
+    "other": 1,
+}
```
VascoSch92, line 386: use Typing.final with the right type here. Thanks

VascoSch92: ah ok it is a test :-)

sleeyax: Yes, exactly, it's a test. Following the project's existing code patterns it's untyped.

enyst:
@openhands Do a /codereview on this PR please. Check out also the linked PR from openhands/openhands so you understand the whole, and review both. Post your feedback using gh api and enyst token.

enyst (approved):
🟢 **Good taste**

LGTM. I reviewed this together with OpenHands/OpenHands#16124. The catalog is the right owner for the taxonomy; the generator validates unknown and conflicting values, preserves the unrelated plugin-category semantics, gives every emitted skill a deterministic fallback, and keeps the generated JS/types/docs aligned. The generator was a clean no-op locally and the category smoke check passed.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 LOW
The change extends a generated public data contract without changing runtime execution or security-sensitive paths. Validation and fallback behavior limit malformed-catalog risk.

**VERDICT:**
✅ **Worth merging**: The producer contract is coherent and ready for the consumer PR.

**KEY INSIGHT:**
The taxonomy has one validated source of truth and reaches consumers through generated artifacts rather than duplicated frontend constants.

_This review was generated by an AI agent (OpenHands) on behalf of enyst._

PR #432, opened 2026-07-31
feat: add the production interface manifest

### LINKED ISSUE #421: Ship the extension-owned Automation interface
## Context

OpenHands/extensions is the sole source of automation-specific presentation, setup, routing, and template information. This issue promotes the reference fixtures into the production Automation interface after the generic Canvas and runtime contracts are ready.

## Acceptance criteria

* Production manifests own all automation routes, navigation, copy, forms, template definitions, validation rules, and action mappings; scheduled and GitHub-event setup work directly with an assisted fallback where required.
* Existing setup, list, detail, dashboard, templates, edit, and import/export behavior remains available through the manifest-driven interface.

## Dependencies

* Blocked by [OpenHands/OpenHands#15783](<https://github.com/OpenHands/OpenHands/issues/15783>).
* Blocked by [OpenHands/OpenHands#15760](<https://github.com/OpenHands/OpenHands/issues/15760>).
* Blocked by OpenHands/OpenHands#15466.
* Blocked by OpenHands/automation#262.
* Blocked by OpenHands/automation#56.
* Blocked by OpenHands/automation#76.

hieptl:

---

## Why

Promotes extension-owned data into the production Automation interface. The merged setup contract (#423) deliberately modeled only the per-entry `setup` block, so every domain-level fact — routes, navigation, page copy, the settable attributes, the import/export envelope, endpoint paths, the proven/responder groupings — still lived hardcoded in Agent Canvas. This PR gives those facts one extension-owned home, stated once at the domain level rather than eight times per entry, and adds the assisted fallback the acceptance criteria require for deployments that cannot run direct setup.

## Summary

- Add `automations/interface.json` (the production interface manifest) and `automations/interface.schema.json` (its strict contract), exported as `AUTOMATION_INTERFACE` with full typings from `./automations` and the root entry, plus package `exports` entries for both files.
- Allow an optional `message` on `mode:"direct"` setup entries — the seed for the fallback conversation offered when a deployment's capabilities preclude the direct path — and add fallback messages to `github-pr-reviewer` and `github-repo-monitor`.
- Document the interface manifest in `automations/README.md` (ownership rules, trust boundary, deviation-table updates) and add `tests/test_interface_manifest.py` (12 tests: schema validity, manifest conformance, mutation rejections, catalog cross-references, fallback messages, node export smoke).

## Issue Number

Resolves https://github.com/OpenHands/extensions/issues/421

## How to Test

1. `uv sync --group test`
2. `uv run --group test pytest tests/ -q` — 512 passed, 6 skipped, including the new `tests/test_interface_manifest.py`.
3. `node --input-type=module -e "import('./automations/index.js').then(m => console.log(m.AUTOMATION_INTERFACE.routes))"` prints the three routes.
4. `npm run build:automations` — `automations/catalog-index.js` is unchanged (the two manifest edits are content-only).

## Video/Screenshots

Not applicable — data/contract change; the pytest run above is the verification.

## Notes

- No version bump in this PR: release-please owns it.
- Agent Canvas consumes `AUTOMATION_INTERFACE` defensively (absent → host defaults), so this releases independently of the Canvas PR.
- The contract fixtures pin `/v1/preset/prompt`; a future change to `endpoints.createPrompt` must regenerate them in the same release (noted in the README section).


automations/interface.schema.json
```diff
@@ -0,0 +1,244 @@
...
+      "description": "Catalog entries the list page surfaces as proven. Each id must resolve to an automations/catalog entry; the host's admission checks that, mirroring how requires.integrations ids are checked by the catalog tests.",
+      "$ref": "#/$defs/slugList"
+    },
+    "responderIntegrationIds": {
+      "description": "Integrations whose automations are treated as continuously-polling responders and get the deployment-choice dialog.",
+      "$ref": "#/$defs/slugList"
+    }
+  },
+  "$defs": {
+    "copy": {
```
enyst, line 168: Does this mean "description" of an automation? If so, maybe we could call it description?

enyst: or simply "text"?

automations/interface.schema.json
```diff
@@ -0,0 +1,244 @@
...
+      ]
+    },
+    "idEndpoint": {
+      "description": "An endpoint path with exactly one {id} substitution and no other braces.",
+      "allOf": [
+        { "$ref": "#/$defs/endpointPath" },
+        { "pattern": "^[^{}]*\\{id\\}[^{}]*$" }
+      ]
+    },
+    "editField": {
```
enyst: Are these... properties or attributes of an automation? Or "variables" for the user to fill in or to select?

If so, could we name this "attributes" or "variables"?

It's a UI-specific concern that it happens to be represented on agent-canvas side by an edit field. If the agent uses this interface via API, "editField" seems a bit of an awkward fit

automations/interface.schema.json
```diff
@@ -0,0 +1,244 @@
...
+        }
+      }
+    },
+    "docsUrl": {
+      "description": "The automations documentation link. Prefix-pinned so this cannot become an arbitrary link.",
+      "type": "string",
+      "pattern": "^https://docs\\.openhands\\.dev/[A-Za-z0-9/._-]*$"
+    },
+    "edit": {
+      "description": "The edit dialog: which Automation properties it offers, keyed by the property each field edits. The closed key set is the editable surface of the runtime model; schedule is a semantic type whose frequency/weekday/time composite the host owns, like the setup form's cron type.",
```
enyst: From what I'm reading, "edit" and "editable" means "input", right? Input from the user. What if we say, idk, like:
```
"variables": {
  "description": "The input for an Automation: which variables it offers, keyed by the variable the user or agent can send. The closed key set is the input surface of the runtime model; schedule ..."
```

automations/interface.schema.json
```diff
@@ -0,0 +1,244 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/OpenHands/extensions/automations/interface.schema.json",
+  "title": "AutomationInterfaceManifest",
+  "description": "Strict schema for automations/interface.json. Mirrors the AutomationInterfaceManifest types in automations/index.d.ts. The catalog states what varies per automation; this file states the domain-level facts of the production Automation interface - routes, navigation, page-identity copy, the edit form, the import/export envelope, the service-relative endpoints, and the featured and responder id lists. The host renders it; it is data, not code - no markup in copy, no free-form URL, no host name, method, or credential.",
```
enyst: This sounds great, just the "edit form" is not domain-level fact of the Automation interface, it's a specific choice of agent-canvas UI - to render it to a user as an edit form.

Unless I'm missing something, for the agents attempting to use the interface, this is simply input to set, or variables to set.

enyst (commented):
Thank you for this! It clarifies a lot and seems to allow us to play with Automations better, even design a new UI for them on the basis of this lot of structured information.

I left a few comments inline, basically on the "edit fields": those don't sound like domain-level Automation definition, they're particular UI-specific, that is, they belong on agent-canvas side. Agent-canvas can choose edit fields for receiving input. But the Automation is not made of edit fields, it seems to me, it would be made of inputs, or variables.

If I may put this differently, if I take a step back and imagine the agents working with Automations (they don't necessarily use canvas), it would nice for them to have a clean and explicit domain structure. That will include inputs, it seems to me.

automations/interface.schema.json
hieptl: The schema's top-level description—and the corresponding text in `automations/index.d.ts` and `automations/README.md`—now refers to **"the settable attributes of an automation"** instead of **"the edit form."** This makes the contract rendering-agnostic, without referring to any specific UI or presentation.

Thank you very much! 🙏

automations/interface.schema.json
hieptl: Agreed. I reframed it to represent the input surface exactly as discussed, with one terminology change: I used **attributes** (your other suggestion) instead of **variables**. In this package, **variables** already refers to the setup form's `{{form.<name>}}` template variables, whereas these represent something different—the runtime model's own properties, which can be updated on an existing automation.

The block is now a top-level `attributes` map keyed by the runtime model property. It is described as the input surface, while leaving the presentation up to the client—for example, Canvas may render an edit dialog, whereas an agent may provide the values through the API. I also moved the previous dialog title into `pages`, alongside the rest of the page identity copy.

Thank you very much! 🙏

automations/interface.schema.json
hieptl: Since these represent attributes of an automation—the keys are the runtime model properties themselves—I renamed the `$defs.editField` definition to `attribute` and added the following description: *"How one settable attribute is offered: a semantic input type, user-visible copy, and value constraints."*

I also updated the corresponding TypeScript types for consistency:

* `AutomationEditField*` → `AutomationAttribute*`
* `AutomationEditableProperty` → `AutomationAttributeName`
* `AutomationInterfaceEdit` → `AutomationInterfaceAttributes`

These changes were applied in both `automations/index.d.ts` and the root `index.d.ts` re-exports.

Thank you very much! 🙏

automations/interface.schema.json
hieptl: I don't think **description** would be the right name here. This field isn't an automation description; it's the shared string type for every user-visible string in the manifest (for example, sidebar labels, command menu titles and keywords, page titles and subtitles, attribute labels, and help text), all of which follow the same no-markup trust rule. Since it covers much more than descriptions, naming it **description** would be misleading.

I'd also prefer to keep **copy** rather than **text** for consistency. The already-released `catalog.schema.json` (0.12.0) defines an identically named `copy` definition with the same wording, and the two sibling schemas are intentionally designed to share a common vocabulary. Renaming it here would introduce an unnecessary divergence, while renaming it there would require changing a published contract.

Additionally, **copy** is the standard UX writing term for user-visible strings that the host renders, which makes it an appropriate and well-established name for this shared definition.

Thank you very much! 🙏

hieptl:
Hello @enyst,

I've updated the code based on your feedback. When you have a chance, could you please take another look at the pull request?

Thank you very much! 🙏

neubig (approved):
LGTM, thanks!

enyst (approved):
I'm curious how easy will the agent work with it now. We'll see 😅 

Thank you!

PR #453, opened 2026-08-05
fix(skills): route "add a skill" to skill-creator

### LINKED ISSUE #454: [Bug]: Saying "add a skill" didn't trigger the add-skill skill
### Operating System

macOS

### Installation Method

npm (`npx @openhands/agent-canvas` or `npm install -g @openhands/agent-canvas`)

### Agent Canvas Version

1.9.0

### Bug Description

Image

Agent created a skill in `.agents/` locally instead of creating a global skill :(

### Steps to Reproduce

tell the agent to add a skill

### Actual Behavior

it creates a local skill

### Expected Behavior

it creates a global skill in .openhands/

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

VascoSch92:
## Problem

Per OpenHands/extensions#454, saying **"add a skill"** did not trigger the intended skill. The request routed toward `add-skill` (which imports an existing skill from a GitHub URL and installs it locally under `.agents/skills/`) instead of `skill-creator` (which is the authoring skill).

Root cause is a trigger-vocabulary collision:
- `skill-creator`'s description matched `create`/`write` but **not** the word `add`.
- `add-skill` is named *add*-skill and claimed `add a skill` in its description.

So the user's exact verb ("add") pulled the request to the wrong skill.

Fixes https://github.com/OpenHands/extensions/issues/454.

## Change (minimal — two frontmatter descriptions only)

- **skill-creator**: add `"add a skill"` / `"make a new skill"` to its triggers, and a one-line note to prefer it (over `add-skill`) when authoring from scratch.
- **add-skill**: scope its description to *importing from a GitHub URL* and point create-from-scratch requests at `skill-creator`, so the two no longer compete for "add a skill".

No logic, script, or file-layout changes.

## Note / out of scope

The issue also expected new skills to land **globally** (e.g. `.openhands/`) rather than locally. That is an Agent Canvas placement concern, not a trigger-routing one, and is intentionally left out of this minimal fix.

Refs OpenHands/extensions#454


skills/add-skill/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: add-skill
-description: Add an external skill from a GitHub repository to the current workspace. Use when users want to import, install, or add a skill from a GitHub URL (e.g., `/add-skill https://github.com/OpenHands/extensions/tree/main/skills/codereview` or "add the codereview skill from https://github.com/OpenHands/extensions/"). Handles fetching the skill files and placing them in .agents/skills/.
+description: Import an existing skill from a GitHub repository URL into the current workspace. Use only when the user provides or references a GitHub URL/repo to fetch from (e.g., `/add-skill https://github.com/OpenHands/extensions/tree/main/skills/codereview` or "add the codereview skill from https://github.com/OpenHands/extensions/"). Handles fetching the skill files and placing them in .agents/skills/. This does not author new skills — to create a new skill from scratch (no source URL), use the skill-creator skill instead.
```
enyst, line 3: What if we named it `import-skill` instead of add?

enyst (approved):

skills/add-skill/SKILL.md
VascoSch92: It is a good thing. However i don't know if it will solve the problem :-) but better naming is for sure something we want in the future

VascoSch92:
@Openhands can you fix the tests?

VascoSch92:
Fixed the failing test in 5ffa45f.

The `test_skills_catalog` check failed because the generated `skills/index.js` was out of date — it still contained the old `add-skill` / `skill-creator` descriptions while the SKILL.md frontmatter had been updated. I regenerated it with `node scripts/build-skills-catalog.mjs`; the only changes are the two edited descriptions. Full suite now passes locally (`uv run --group test pytest tests/` → 539 passed, 6 skipped).

_This comment was created by an AI agent (OpenHands) on behalf of the user._

PR #456, opened 2026-08-05
feat: add openhands-enterprise-troubleshooting skill for OHE self-hosted debugging

### LINKED ISSUE #527: Skill for diagnosing self-hosted OpenHands Enterprise installs
### Desired Behavior

An agent-runnable skill for diagnosing self-hosted OpenHands Enterprise (Replicated VM-based) installs, usable by customers self-serving, by an FDE during a support call, or as a first-response step before escalating to the platform team.

It should triage the eight failure modes that recur across real installs — sandbox startup timeouts, git provider auth, certificate errors, LLM connectivity, Keycloak login, Replicated Admin Console access, stuck upgrades, and resource exhaustion — from a described symptom, pasted log output, or a support bundle. Recovery should validate each step before the next rather than only at the end, so a failed step is caught before a later one buries the evidence.

Support bundles need particular attention: the signal is spread across `analysis.json`, per-namespace pod objects, and node metrics, so the skill should summarise a bundle offline and rank findings into a most likely root cause rather than listing everything it sees. Where the bundle's own tooling misleads — `warn`-severity analyzers rendering under a `FAIL` heading, and per-object analyzers collapsing to one arbitrary example per family — the skill should say so, because both invite a confident wrong answer.

When an issue cannot be resolved, it should produce an escalation handoff stating what was tried, what the logs show, and the likely root cause.

### Acceptance Criteria

- [ ] Triage covers all eight failure modes, from symptom, pasted logs, or a bundle.
- [ ] Guided recovery validates each step before proceeding, and requires agreement for destructive steps.
- [ ] Support bundle generation is documented, and bundle output is summarised into a ranked root cause.
- [ ] The two analyzer traps (`warn` under `FAIL`, `[xN]` family collapse) are documented.
- [ ] An escalation handoff template is included.
- [ ] Commands are checked for accuracy rather than assumed: `kubectl` invocations parse, `jq` programs compile, and no code block references an undefined shell variable.
- [ ] The bundle triage script remains a verbatim copy of its source, verified by checksum.
- [ ] `uv run pytest -q` passes and `npm run build:skills` regenerates the catalogue with no diff.

### Context

Implemented by PR #456. Derived from Linear PLTF-2910, which tracks the same work internally.

Known limitation to resolve before this can be trusted in the field: the skill hard-codes namespaces, label selectors, deployment names, and secret names that cannot be verified offline. A wrong selector returns `No resources found` and exits 0, which reads as "healthy" rather than "wrong query", so these need checking against a real install.

_This issue was created by an AI agent (OpenHands) on behalf of @ai.vong._

jpelletier1:

---

## Why

Diagnosing a self-hosted OpenHands Enterprise install currently depends on knowledge that lives with individual engineers. The same eight failures recur — sandbox startup timeouts, git provider auth, certificates, LLM connectivity, Keycloak, the Replicated Admin Console, stuck upgrades, resource exhaustion — and each investigation restarts from nothing. Support bundles make this worse rather than better: they are large, and the signal is spread across `analysis.json`, per-namespace pod objects, and node metrics, so reading one by hand is slow and easy to get wrong.

This skill gives customers, FDEs, and first response a repeatable path from symptom to either a fix or an escalation that the platform team can act on without a second round of questions.

## Summary

- Adds the `openhands-enterprise-troubleshooting` skill: triage for the eight common failure modes, guided recovery that validates each step before the next, support bundle generation, and an escalation handoff template.
- Adds `references/diagnostics.md` (per-failure-mode commands) and `references/support-bundle-analysis.md` (bundle layout, log triage, and how the bundle's own analyzers mislead).
- Adds `scripts/bundle_triage.py`, which summarises a bundle offline — meta, analyzers, pods, restarts, top, alloc, events.
- Documents how to rank findings into a most-likely root cause, including two traps in the tooling: `warn`-severity analyzers print under a `FAIL` heading, and per-object analyzers show one arbitrary example per family, so `[x12]` is a count and not a worst case.
- Registers the skill in the marketplace catalogue and skills index.

## Issue Number

Closes #527

## How to Test

```bash
uv run pytest -q            # 786 passed, 14 skipped
npm run build:skills        # catalogue regenerates with no diff
```

To exercise the script against a real bundle:

```bash
python skills/openhands-enterprise-troubleshooting/scripts/bundle_triage.py <bundle-dir>
python skills/openhands-enterprise-troubleshooting/scripts/bundle_triage.py <bundle-dir> --section analyzers,pods
```

Read `SKILL.md` end to end — most of the change is prose, and its accuracy is the thing under review.

## Video/Screenshots

n/a — no user interface.

## Notes

`scripts/bundle_triage.py` is a verbatim copy of an existing internal script and is deliberately unmodified, so that improvements land at the source rather than diverging here. It has known rough edges: it raises on bundles missing `node-metrics` node info, on analyzer entries without an insight, and on nodes without pod capacity. Those are worth fixing upstream and then re-syncing.

Command accuracy was checked rather than assumed: every `kubectl` invocation was parsed and run against an unreachable cluster to surface flag and subcommand errors, every single-line `jq` program was compiled, and code blocks were scanned for undefined shell variables. That found two genuine errors, both fixed here — `kc.sh health` is not a Keycloak subcommand (health is HTTP on the management port, and only when enabled), and the database connectivity check referenced `$DB_HOST`/`$DB_PORT`, which are set nowhere and, inside single quotes, expanded in the pod rather than the shell.

The bundle paths documented in `references/` were verified against real bundles in earlier sessions; those bundles are no longer on the machine, so they were not re-verified in this pass.

---
_This description was updated by an AI agent (OpenHands) on behalf of @ai.vong._


aivong-openhands:
## Added support bundle analysis

This skill covers generating a support bundle and reading one: the real on-disk layout, how to reconstruct cluster state from it, and a script that does the mechanical first pass.

### What was added

- **`references/support-bundle-analysis.md`** — the real bundle layout: a `kubectl` → bundle-path map, reconstructing pod state from `cluster-resources/pods/<ns>.json`, the three independent OOM sources and why they disagree, the node-reboot signature, log triage, and the gaps that produce convincing false negatives (an empty events file ≠ nothing happened; `***HIDDEN***` means redacted, not unset; log filenames are *container* names, so init-container failures like `migrate-db` are invisible to `kubectl logs <pod>`; mtimes reflect extraction time, not capture time).
- **`scripts/bundle_triage.py`** — offline triage in one command: cluster meta, analyzer verdicts, pod table, OOM/restart scan, `top` equivalent, allocatable headroom, events. Standard library only.
- Cross-references so each live-cluster failure mode in `diagnostics.md` maps to its offline equivalent.
- A note that bundles carry potentially sensitive data and that the redactor is pattern-based, so it is not a guarantee.

Two things the tool deliberately won't do. **It doesn't pretend to diagnose** — it reports what the objects say and leaves to the reader which of those observations explains the user's symptom. And **a clean run doesn't claim health**: the guide states what a clean run actually rules out, notes that the script reads no application logs, and points to where to look next. Log analysis is deliberately left to the agent rather than the script: a fixed matcher only finds the patterns it was given and cannot tell you what it missed, whereas the guide's method — judge the format, cluster by message shape, bucket by time, read the burst — is exactly the kind of judgement an agent applies better than a regex. Nobody sends a support bundle because things are working, so "no findings" is the hard case, not the good one — and telling a broken customer their bundle looks healthy is the outcome most worth designing against.

### Accuracy

Every command and object name is verified against the Replicated docs and the OpenHands-Cloud charts rather than inferred, because a plausible-looking wrong command costs a support engineer more than no command at all. What that check settled:

- **Bundle collection** is `sudo ./openhands support-bundle`, with a `kubectl support-bundle` fallback for pre-1.17.0. The vendor-side `replicated` CLI is not present on a customer VM.
- **Namespaces**: the Admin Console runs in `kotsadm` and the operator in `embedded-cluster`, and `kubectl` needs `sudo ./openhands shell` first.
- **Rollbacks** go through the Admin Console, with a warning against hand-editing Helm state.
- **Git provider credentials** live in per-provider secrets — `github-app`, `gitlab-app`, `bitbucket-data-center-app`, `azure-devops-app`.
- **Sandboxes** are created on demand by `runtime-api` as individual `runtime-*` pods; there is no sandbox deployment or `app=sandbox` label.
- **The container runtime** is the `sysbox-runc` RuntimeClass, registered per node by a DaemonSet.
- **The bundle layout** is what `troubleshoot.sh` actually produces — `cluster-resources/pods/<ns>.json` and container-named logs, not `pod-status.json` or `pod-logs/*.log`.

**Git provider coverage is provider-neutral.** Bitbucket Data Center and Azure DevOps are first-class providers alongside GitHub and GitLab, each with its own secret and template. Both are self-hosted, so their host comes from configuration rather than a fixed domain, and an apparent auth failure is often really network or TLS — so the reader is prompted to establish which provider the user is on first.

The credential check reports key names and byte lengths rather than secret values, since a present-but-empty key is the failure worth finding and Helm renders empty values into valid-looking Secrets.

### Validation

The interpretation rules were checked against real Embedded Cluster bundles, including unhealthy ones (failed migrations, a crash-looping operator, Pending pods, sandbox-inflated pod counts). Healthy bundles were used only to learn what normal looks like and to catch false positives; every claim about a failure mode was checked against a bundle actually exhibiting it, or is marked unproven.

Four cases where the obvious reading of the data is wrong, and which the guide therefore spells out: node conditions whose healthy value is `True` (`ContainerdHasNoDeprecations` is not a fault); init-container requests, whose true cost is `max(sum(regular), max(init))` rather than the sum of both, with native sidecars (`restartPolicy: Always`) belonging in the running sum; analyzer results that must be kept apart by subsystem, so a real failure is not read as an instance of an unrelated one; and recovered terminations, which are distinguishable from active ones only by their age.

`references/support-bundle-analysis.md` carries a **Validation status** section recording what is verified against real data and what isn't. Still fixture-only: an actively OOMing container, a truly unschedulable pod, the `EXCEEDS ALLOCATABLE` path, the host-reboot heuristic, and native sidecars.

### The triage script is a verbatim copy

`bundle_triage.py` is byte-identical to the script it is taken from, deliberately.

The guide covers scheduling constraints as their own failure mode: a `Pending` pod whose `nodeSelector` or `affinity` nothing satisfies is stuck regardless of how much capacity the cluster has, and the allocation numbers will not explain it. Both halves of the skill say to read the scheduler's own reason first — `describe pod` on a live cluster, the pod spec and node labels offline.

The skill ships the `.plugin/plugin.json` and vendor symlinks the marketplace tests require, and the generated catalog is current. Full suite: **786 passed**.

---
_This comment was created by an AI agent (OpenHands) on behalf of @ai.vong._

aivong-openhands:
## Closing three gaps against the requirements

An audit against the original requirements turned up three things the skill did not do. All three are addressed in prose — `bundle_triage.py` is unchanged and still byte-identical to its source.

### Parse and summarize the bundle to highlight the most likely root cause

The script reports observations and deliberately stops short of ranking them, so the ranking needs to live in the guide. A new **Summarizing the Bundle: Most Likely Root Cause** subsection walks the output in the order a finding is likely to be a cause rather than a side effect: place the symptom in time first, read `analysis.json`, then rank live crash loops and OOM above pods that never started, above `Running`-but-not-`Ready`, above node conditions, above resource pressure — which is usually a consequence.

Two ways the analyzer summary misleads are worth having written down, because both invite a confident wrong answer:

- Everything non-passing prints under a `FAIL` heading, `warn` severity included. Some of those are advisory. Check the severity in `analysis.json` before calling one a failure to a customer.
- Per-object analyzers are collapsed into families with a single example. `[x12]` means twelve objects affected and the line shown is an arbitrary one of them, not the worst.

Two further habits the section asks for: prefer the cause nearest the symptom, since a failed `migrate-db` init container and the app pod stuck behind it are one finding rather than two; and when the objects explain nothing, say so as a result — it locates the cause in the application logs, the network path, or outside the cluster, which is more useful than reporting that the bundle looked healthy. The output is a ranked shortlist with its evidence, which drops into the **Likely Root Cause** field of the handoff template.

### Validate each recovery step before proceeding

The workflow verified once, at the end. That is the point at which a half-applied fix has already buried the evidence that would have told you the first step failed. Recovery now states what a step will change and what to expect, checks it before the next one, and stops to re-diagnose when the check disagrees. Destructive steps — restarts, rollbacks, config changes — need the user's agreement, and get recorded as they happen so the handoff can say what was touched.

### Raw log output pasted into the conversation

Log triage was written against files, which is not how logs usually arrive. The three moves transfer to pasted text unchanged; the hazards do not, so the two that matter are named. A pasted excerpt is a *selection*, made by someone who already had a theory — the cause often sits in the lines just before what you were handed, so ask for the surrounding context rather than reasoning from the fragment. And it comes without a filename, so the container that produced it is unknown: the same stack trace from an init container and from the main container mean different things. The diagnostic workflow now routes all three input types — described symptom, bundle, pasted logs — from the top of `SKILL.md`.

Full suite: **786 passed**. `bundle_triage.py` unmodified.

---
_This comment was created by an AI agent (OpenHands) on behalf of @ai.vong._

skills/openhands-enterprise-troubleshooting/SKILL.md
```diff
@@ -0,0 +1,324 @@
...
+
+# Check resource usage
+kubectl top nodes
+kubectl top pods -n openhands
+
+# Check certificate expiry
+echo | openssl s_client -connect HOST:443 2>/dev/null | openssl x509 -noout -dates
+
+# Check Replicated operator
+kubectl get pods -n replicated
```
aivong-openhands: Not sure if there is a replicated namespace 🤔

aivong-openhands: Confirmed there isn't one

aivong-openhands: Confirmed and fixed in 2e05efe. You were right — validating against six real support bundles and a live v0.58.0 instance showed no `replicated` namespace and no `app=replicated-operator` label anywhere. It is kURL-era topology that Embedded Cluster does not use.

The check now points at where those components actually live: the Admin Console in `kotsadm`, the operator in `embedded-cluster`, and the Replicated SDK in `openhands` under `app.kubernetes.io/name=replicated`.

Worth flagging why this class of bug matters more than it looks: a wrong namespace or selector does not error. `kubectl get pods -n replicated` prints `No resources found` and exits 0, so an agent reads a broken query as a healthy component. Live testing found three selectors failing exactly this way, all now corrected.

_This reply was posted by an AI agent (OpenHands) on behalf of @rbren._

aivong-openhands: Fixed namespace in https://github.com/OpenHands/extensions/pull/456/changes/2e05efeed166a8bcac35e2e54e0c23c57b2c132e

dylan-openhands (approved):

PR #483, opened 2026-08-18
feat(github-pr-reviewer): ship the reviewer script as a catalog bundle

VascoSch92:
Stacked on #472 — review that one first; this branch is #472 plus one commit.

## What this changes

`automations/catalog/github-pr-reviewer/manifest.json` created a **preset-prompt** automation, so the declarative path picked up none of the machinery `skills/github-pr-reviewer/scripts/main.py` owns — #472 says as much in its own "Not changed" section. The agent-driven path packs that script by hand (SKILL.md steps 6–8: substitute five constants, `tar -czf`, `POST /v1/uploads`, `POST /v1`). This makes the catalog entry do the same thing declaratively.

It depends on `OpenHands/automation#346`, which adds `template` provenance, `/v1` preflight and the `customTarball` capability to the raw create path — the three things the bundle flow needs and the preset paths already had.

### `setup.bundle`, the alternative to `setup.prompt`

```jsonc
"bundle": {
  "version": "1.0.0",
  "entrypoint": "python3 main.py",
  "timeout": 600,
  "files": { "main.py": "skills/github-pr-reviewer/scripts/main.py" },
  "config": {
    "repos": ["{{form.repository}}"],
    "trigger_label": "{{form.triggerLabel}}",
    "review_tone": "{{form.reviewTone}}"
  }
}
```

Direct mode now produces exactly one of `prompt` or `bundle`, enforced by the schema. Everything else in the create request is still restated from the form and therefore still absent: only `config` is declared, because only the entry knows which key of its own script each field fills. That is the bundle's `prompt`.

`requires.features` becomes `["customTarball"]`. It declares no `repos`: the raw create endpoint has no such field — verified, it is a hard `extra_forbidden` 422 — and a bundle fetches what it needs itself.

### One copy of the script

`files` names repository paths rather than carrying contents, so the script shipped by both the skill and the catalog does not exist twice and cannot drift. `npm run build:automations` inlines the contents into the generated `automations/bundle-index.js`, exposed as `getAutomationBundleFiles(id)` — a host packing the archive has the published package, not this repository. The build refuses a path outside `skills/` or `automations/`, and the schema refuses one with a scheme or a traversal.

### `config.json` instead of rewriting Python

The five constants stay exactly where #472 put them and are now the *defaults*; `load_config()` reads a `config.json` shipped beside the script and overrides them. That is what lets the catalog pack an unmodified copy — a declarative host cannot rewrite Python, and the catalog schema admits data, not code. The agent-driven path substitutes the constants and ships no `config.json`, so the two never meet.

Types are checked rather than trusted: a `repos` that arrives as a string would otherwise be polled one character at a time.

### `interface.json`

Two endpoints the host now calls: `createBundle: "/v1"` and `uploads: "/v1/uploads"`. Both are service-relative, as the block requires.

## Testing

- `uv run --group test pytest tests/` — **543 passed, 12 skipped** (541/6 on #472; +2 tests, and the two new per-entry tests skip for the eight non-bundle entries).
- `python -m pytest tests/` in `skills/github-pr-reviewer/` — **36 passed** (26 on #472; 10 new for the config loader: partial config, unknown keys ignored, a string where a list belongs, an empty or non-string repo list, malformed JSON, a JSON array).
- `sync_extensions.py --check`, `sync_openhands_sdk_skill.py --check` — clean. `npm run build:skills` regenerated `skills/index.js` for the SKILL.md note; `npm run build:automations` regenerated both indexes.
- The contract fixture is rewritten for the bundle path, and the two 422 bodies in it were **re-verified against a live service** running `OpenHands/automation#346` rather than hand-written.

New contract tests: every `files` path exists and matches what the generated index inlined; the entrypoint names a packed file; every `{{form.*}}` in `config` names a field the form collects.

## Fixtures worth a look

- `capabilities.json` gains `customTarball` on both ready deployments — it is a static feature of the service, so every ready deployment reports it — plus a fourth shape, `noCustomTarball`, standing for a service that predates bundle support. That is what makes the capability gate testable: this entry is now `blockedBy: ["noCustomTarball", "notReady"]`.
- `short-repo-url-without-provider-is-rejected` was about `repos[].provider`, which the bundle path does not have. It is replaced by `repos-on-the-raw-create-request-is-rejected` — the same class of mapping bug in the shape this path can actually hit.

## Worth a second opinion

- **`bundle.version` is entry-owned**, not the package version, which moves for reasons unrelated to this entry. It is what tells an already-enabled deployment that what it installed is stale. Nothing consumes that comparison yet.
- **The form still collects one repository**, though #472's script polls many. `config.repos` renders a single-element list. Multiple repositories need either a repeatable field type or a comma-split, and both are host-side changes.
- **`setup.message` is unchanged** and still describes the fallback conversation. It is now also what a `noCustomTarball` deployment falls back to.


VascoSch92:
Pushed two fixes found by installing this on a live stack.

**A clone URL is not a repository name.** `https://github.com/owner/repo` became `/repos/https://github.com/owner/repo`, which GitHub answers with a 404, which the script reported as *"Repository … is not accessible with the current token"* — pointing at the token when the repository was public and reachable. `normalize_repo` reads the ways a repository gets written down (name, https URL, ssh remote, `.git` suffix, trailing slash) and rejects anything else *by naming the value it could not read*. It runs per repository inside the poll loop, so a bad one fails on its own and the completion callback still fires. 10 tests.

**The entry collected one repository while the script polls many.** A `repo-picker` may now declare `multiple: true`, and then its value is a list. That needed one rule, not a special case: a placeholder that is the *whole* value resolves to that value's own type, so `"repos": "{{form.repositories}}"` produces an array while the same placeholder inside a sentence still reads as text. On the preset path the list becomes one `repos[]` entry each. The automation is named after the single repository when there is one and after the count when there are several — a list of names does not fit a name.

New `several-repositories` contract scenario pins the two-repository create body; the host reproduces it in `OpenHands/OpenHands#16680`.

554 passed / 12 skipped, 46 in the skill.

neubig (approved):

PR #486, opened 2026-08-19
feat: add Knowl integration

williamttruong:
HUMAN:
Ran `npx -y @dat999zx/knowl@5.6.0 serve` in a fresh Windows cmd.exe shell. The MCP server started and reported ready in 0ms over stdio. I'm a co-founder of Knowl, so treat the description as a claim to check rather than take on faith.

---

## Why

The catalog's only memory entry is `memory`, the reference `@modelcontextprotocol/server-memory`, which is a key-value store scoped to conversations. There is no entry for project memory: the decisions, constraints and architecture of a codebase, and which of them are still true.

Knowl fills that gap. It resolves conflicts at **write** time: a fact whose subject and relation match one already stored retires the earlier record, so the stale value never becomes a retrieval candidate. In MemoryAgentBench's own harness on FactConsolidation-SH at 262k, that mechanism is worth 89.0 with supersession on against 73.0 with it off.

## Summary

- Adds `integrations/catalog/knowl.json`
- Regenerates `integrations/catalog-index.js` via `npm run build:integrations`
- stdio MCP transport, `npx -y @dat999zx/knowl serve`
- `auth.strategy: "none"`, so it is local-first with no credentials and nothing to configure in the install modal

## Issue Number

n/a

## How to Test

```bash
npx -y @dat999zx/knowl serve
```

Speaks MCP over stdio and registers its tools. `npx -y @dat999zx/knowl --help` lists the CLI surface. Package is published at `@dat999zx/knowl` (5.6.0), source at https://github.com/dat999zx/knowl. Verified on Windows via `cmd.exe`, where the `npx` form works without the `knowl.cmd` shim the project's own README mentions.

The entry matches the `IntegrationCatalogEntry` shape in `integrations/index.d.ts`, with required `id`, `name`, `description`, `docsUrl` and `connectionOptions` all present. Categories reuse existing vocabulary (`Knowledge base` from notion, `Engineering` from github and linear) rather than introducing new values.

Not tested through the OpenHands install modal, since I don't have that environment.

## Video/Screenshots

n/a. Catalog metadata only, no UI change beyond the new entry appearing in the list.

## Notes

Disclosure: I'm a co-founder of Knowl. Happy to adjust the description, categories or `iconBg` to house style, or drop the benchmark reference from this body if you'd rather the catalog stay claim-free.

### DIFF
--- a/integrations/catalog-index.js
+++ b/integrations/catalog-index.js
@@ -67,17 +67,18 @@ import entry61 from "./catalog/firecrawl.json" with { type: "json" };
 import entry62 from "./catalog/git.json" with { type: "json" };
 import entry63 from "./catalog/huggingface.json" with { type: "json" };
 import entry64 from "./catalog/kagi.json" with { type: "json" };
-import entry65 from "./catalog/memory.json" with { type: "json" };
-import entry66 from "./catalog/mongodb.json" with { type: "json" };
-import entry67 from "./catalog/neon.json" with { type: "json" };
-import entry68 from "./catalog/obsidian.json" with { type: "json" };
-import entry69 from "./catalog/paypal.json" with { type: "json" };
-import entry70 from "./catalog/playwright.json" with { type: "json" };
-import entry71 from "./catalog/redis.json" with { type: "json" };
-import entry72 from "./catalog/resend.json" with { type: "json" };
-import entry73 from "./catalog/sequential-thinking.json" with { type: "json" };
-import entry74 from "./catalog/superhuman-mail.json" with { type: "json" };
-import entry75 from "./catalog/time.json" with { type: "json" };
+import entry65 from "./catalog/knowl.json" with { type: "json" };
+import entry66 from "./catalog/memory.json" with { type: "json" };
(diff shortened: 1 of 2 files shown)


VascoSch92:
@williamttruong 

Thanks for the submission, and for the upfront disclosure.

There are a couple of problems.

The entry declares auth.strategy: "none" with "No credentials or configuration required", but knowl serve starts regardless of project state and then returns Knowl project is not initialized. Run "knowl init" first. from every tool call and resource read (the server holds an initError and surfaces it through registerTools/registerResources). Installed from the catalog modal, that advertises 27 tools into the agent's context and fails on all of them, since there is no step in that flow that runs knowl init.

There is also an architectural mismatch worth flagging: knowl init adds .knowl/ to .gitignore, and OpenHands conversations get a fresh runtime and a fresh checkout, so the store is empty at the start of every session. Cross-session memory is exactly the part that does not carry over.

Closing for now. 

If you are interested in adding that to openhands, i will suggest you to open first an issue such that we can discuss about that.

PR #488, opened 2026-08-20
feat(github-agents-md-maintainer): keep AGENTS.md current on a schedule

VascoSch92:
## What

A second automation in the shape of `github-issue-to-pr`: a scheduled agent that keeps `AGENTS.md` current.

- `skills/github-agents-md-maintainer/` - `SKILL.md`, `scripts/main.py`, `references/state-schema.md`, README, plugin manifest
- `automations/catalog/github-agents-md-maintainer/manifest.json` - card metadata plus a `direct` setup block shipping a **bundle** (repositories multi-select, branch prefix, draft-or-ready, cron)
- `marketplaces/openhands-extensions.json`, `tests/test_github_agents_md_maintainer.py` (21 tests)

Weekly by default (`0 9 * * 1`). Per run, per repository: clone the default branch, start a conversation that reads the repository and edits `AGENTS.md`, open a pull request, remove the clone.

## The two rules that keep it quiet

Scheduled documentation automations fail by being noisy, so both failure modes are designed out:

1. **One unit of work is one repository in one ISO week** (`agents-md:2026-W34`). A cron that fires more often than intended, a retried run, or a restarted service all resolve to the same key, so the work happens once. Extra runs inside a week only poll.
2. **A repository whose previous pull request from this automation is still open is skipped**, and the skip is recorded in state. Without this, a repo nobody is merging collects a pull request per week, each editing the same file. Only branches under the configured prefix count as "ours" - someone else's open PR never silences it, and a failed listing does not either (failing closed there would let an API blip skip the week).

A week where `AGENTS.md` is already accurate ends with **no pull request** - the expected outcome most weeks. The agent's reasoning is still recorded in state as `summary`.

## Division of labour

Same as issue-to-pr: the agent reads the repository, edits the file, commits, pushes, and opens the pull request, so it lands as soon as it finishes. The script verifies on GitHub rather than trusting it, and opens the pull request itself when the agent did not - a failed push, a dead conversation, or an agent that simply did not bother still ends with the right outcome. A failed push or `POST /pulls` is retried on the next two runs before it gives up.

Title is `docs: add AGENTS.md` or `docs: update AGENTS.md`, decided by asking GitHub whether the file exists - and an unreadable answer resolves to *update*, because proposing to "add" a file that exists reads worse than the reverse.

## What the agent is told

The prompt does not restate what belongs in an AGENTS.md - those rules are the `agent-memory` skill's, and the prompt follows them: general knowledge only, no task-specific notes, no unverified commands, no secrets, a page or two at most. It is explicitly told to treat an existing file as someone else's writing (correct, add, delete - not rewrite) and that changing nothing is a valid outcome.

The token requirement is *smaller* than issue-to-pr's: contents write and pull request write, **no issue permission**, because this automation never comments on an issue.

## Testing

- `uv run --group test pytest tests/` - 595 passed, 12 skipped
- A network-free end-to-end harness (in my scratchpad, not the repo) drives the real script against a bare git repository with only the GitHub API and agent server faked - 24 checks covering: stale file → update PR pushed to the remote, missing file → add PR, accurate file → no PR, an open PR in flight → week skipped, two runs in one week → one conversation, agent-opened PR → adopted rather than duplicated.
- `python scripts/sync_extensions.py --check` clean, `npm run build` regenerated with no drift.

Not exercised against a live deployment yet.

## Note

The OSS VM runs something comparable; I could not read it this session because `gcloud` auth had expired, so this was built from the issue-to-pr pattern rather than from that script. Worth a look before merge in case the VM version has prompt wording worth adopting.


skills/github-agents-md-maintainer/references/state-schema.md
```diff
@@ -0,0 +1,113 @@
...
+
+The automation maintains a JSON state document **per repository**, persisted
+across runs. It is the source of truth for which weeks have been handled, which
+conversations are still active, and which clones are still on disk.
+
+---
+
+## Storage
+
+**Primary (cloud):** the automation service's KV store, key
```
enyst, line 11: out of curiosity, why is cloud "primary"?

tests/test_github_agents_md_maintainer.py
```diff
@@ -0,0 +1,352 @@
...
+        lambda token, path, params=None: [
+            {"number": 1, "head": {"ref": "feature/something"}},
+            {"number": 2, "head": {"ref": "openhands/agents-md-2026-W33"}},
+            {"number": 3, "head": {"ref": "openhands/issue-42"}},
+        ],
+    )
+
+    in_flight = main._open_pull_requests_from_this_automation("tok", "acme/widget")
+
+    assert [pr["number"] for pr in in_flight] == [2]
```
enyst, line 88: testing private methods directly might not be terribly useful since sure, it works now, but will we notice in the future that the agent changed the test? Or rather, please let me put it this way: unit tests should test units of behavior, and something that changes when the private implementation changes is maybe not a useful signal for correctness

enyst (approved):
Thank you!

PR #494, opened 2026-08-21
fix: use Notion MCP token environment variable

DevinVinson:
## Summary
- configure the Notion MCP integration with `NOTION_TOKEN`, the variable supported by `@notionhq/notion-mcp-server`
- add regression coverage for the credential environment variable

## Validation
- `npm run build:integrations`
- `uv run --group test pytest -q tests/test_catalog_schema.py`
- `python scripts/sync_extensions.py --check`

_This pull request was created by an AI agent (OpenHands) on behalf of Devin Vinson._

### DIFF
--- a/tests/test_catalog_schema.py
+++ b/tests/test_catalog_schema.py
@@ -78,16 +78,6 @@ def test_schema_requires_catalog_and_install_metadata() -> None:
         assert any(field in error.message for error in errors), errors
 
 
-
-def test_notion_mcp_uses_supported_token_environment_variable() -> None:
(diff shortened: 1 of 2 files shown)


tests/test_catalog_schema.py
```diff
@@ -78,6 +78,16 @@ def test_schema_requires_catalog_and_install_metadata() -> None:
         assert any(field in error.message for error in errors), errors
 
 
+
+def test_notion_mcp_uses_supported_token_environment_variable() -> None:
```
enyst: We probably don’t need a test for this 😇

enyst (approved):
Thank you for the investigation and fix!

PR #522, opened 2026-08-28
ci: add issue duplicate checker workflows (parity with sdk/OpenHands)

smolpaws:
HUMAN:
This PR proposes to add a duplicate check for issues, so that they can be automatically linked to potential dupes, or automatically closed.

### What

Adds the **issue duplicate checker** workflows to this repo, matching what `OpenHands/software-agent-sdk` and `OpenHands/OpenHands` already run:

- `issue-duplicate-checker.yml` — on new issues (plus a daily `cron` sweep + manual `workflow_dispatch`), calls the `OpenHands/extensions/plugins/issue-duplicate-checker` action to label likely duplicates `duplicate-candidate` and auto-close aged candidates.
- `remove-duplicate-candidate-label.yml` — clears the `duplicate-candidate` label when a human comments, so real activity un-flags it.

### Why

`extensions` already runs `issue-readiness-check.yml` but has no duplicate detection, unlike the sibling repos. This brings its issue triage to parity.

### Notes

- Files are the repo-agnostic versions from `OpenHands/OpenHands` (parameterized on `${{ github.repository }}`), copied verbatim so behavior matches the other repos.
- This repo **hosts** the `plugins/issue-duplicate-checker` action; the workflow references it by the same pinned commit (`fb020c71…`) the siblings use, i.e. a self-reference at a fixed sha. If you'd prefer a local path reference (`./plugins/issue-duplicate-checker`) here instead, say the word and I'll switch it — I kept it identical to the siblings for consistency.
- **Secrets:** the check job uses `OPENHANDS_API_KEY` + `OPENHANDS_BOT_GITHUB_PAT_PUBLIC` (falls back to `github.token`). Please confirm `OPENHANDS_API_KEY` is available to this repo at the org level before merging.
- Workflows only; no app changes.

Opened at @enyst's request (parity pass across the main repos).

### DIFF
--- a/.github/workflows/issue-duplicate-checker.yml
+++ b/.github/workflows/issue-duplicate-checker.yml
@@ -1,0 +1,92 @@
+---
+name: Issue Duplicate Check via OpenHands Cloud
+
+on:
+    issues:
+        types: [opened]
+    schedule:
+        - cron: 0 9 * * *
+    workflow_dispatch:
+        inputs:
+            mode:
+                description: Which workflow path to run
+                required: true
+                type: choice
+                options:
+                    - smoke-clone
+                    - issue-check
+                    - auto-close
+                default: smoke-clone
+            issue_number:
+                description: Existing issue number to analyze when mode is issue-check
+                required: false
+                type: number
+            close_after_days:
+                description: Days to wait before auto-closing duplicate candidates in auto-close mode
+                required: false
+                type: number
+                default: 3
+
+permissions:
+    contents: read
+    issues: write
+
+jobs:
+    smoke-clone:
+        if: github.event_name == 'workflow_dispatch' && inputs.mode == 'smoke-clone'
+        runs-on: ubuntu-latest
+        timeout-minutes: 5
+        steps:
+            - name: Clone repository
+              env:
+                  REPOSITORY: ${{ github.repository }}
+              run: |
+                  git clone --depth 1 "https://github.com/${REPOSITORY}.git" /tmp/repo-clone
+                  echo "REPO_HEAD=$(git -C /tmp/repo-clone rev-parse --short HEAD)" >> "$GITHUB_ENV"
+
+            - name: Summarize smoke test
+              run: |
+                  {
+                    echo "## Smoke clone completed"
+                    echo
+                    echo "- Repository cloned to /tmp/repo-clone"
+                    echo "- Repository HEAD: ${REPO_HEAD}"
+                  } >> "$GITHUB_STEP_SUMMARY"
+
+    issue-duplicate-check:
+        if: |
+            github.event_name == 'issues' ||
+            (github.event_name == 'workflow_dispatch' && inputs.mode == 'issue-check' && inputs.issue_number != '')
+        runs-on: ubuntu-latest
+        timeout-minutes: 35
+        concurrency:
+            group: issue-duplicate-check-${{ github.repository }}-${{ github.event.issue.number || inputs.issue_number }}
+            cancel-in-progress: false
+        steps:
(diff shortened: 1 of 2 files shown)


enyst (changes requested):
## Code Review — `/codereview`

🟡 **Acceptable** — the two workflow files are correct, but this PR lands them with a self-reference, an unconfirmed secret, and a failing PR-description check, so it should not merge as-is.

### What this PR does

Adds the same `issue-duplicate-checker.yml` + `remove-duplicate-candidate-label.yml` pair that `OpenHands/OpenHands` and `software-agent-sdk` run. The action reference (`plugins/issue-duplicate-checker@fb020c71…`) resolves; I confirmed the `plugins/issue-duplicate-checker/` tree at that sha is unchanged on `main`, so the pin is valid and stable.

### Blocking issue: `OPENHANDS_API_KEY` is not available to this repo

The `issue-check` path hard-fails when `OPENHANDS_API_KEY` is absent. `extensions` has **zero** repo actions secrets (verified via API), and the org-secrets endpoint 403s for my token, so I can't confirm an org-level secret. This is not hypothetical — the identical workflow is red on every `issues` event in `OpenHands/OpenHands` with `Error: openhands-api-key is required for issue-check mode` (run 33203722361). Merging as-is means a guaranteed red check on every new issue here too. **Please confirm the org secret exists (or add it) before merging.**

### Should be fixed: use a local-path self-reference

The PR body flags this: this repo **hosts** the action, and the workflow references it by pinned sha. For a repo referencing an action it contains, `./plugins/issue-duplicate-checker` is strictly better — the workflow then tests the action at the same commit as the PR, and a change to the action can't silently be used by every other repo before this repo has exercised it. I'd ask for the local-path reference here rather than mirroring the siblings; this is the one repo where the self-reference is meaningful, not cosmetic.

### Blocking issue: PR description fails the repo's own validation

CI `Validate PR description` is failing (run 33218008668) with four errors: missing `## Why`, `## Summary`, `## How to Test`, and no linked issue in `## Issue Number`. This is the repo's own gate — the description should be restructured to the PR template. (For what it's worth, that template is exactly why I can't merge a PR whose evidence is "copied verbatim from the siblings" without the secret confirmation.)

### Minor

- The `github.token` fallback (`OPENHANDS_BOT_GITHUB_PAT_PUBLIC || github.token`) means the REST calls to `api.github.com` inside the action run unauthenticated; sibling repos already hit `API rate limit exceeded` on this path (`software-agent-sdk` run 33152648707). Same latent trap here.

### [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟡 MEDIUM — workflow files themselves are low-risk and well-formed, but the PR is missing the secret it depends on, fails its own CI gate, and uses a pinned sha where a self-reference is both available and preferable. Recommend: fix description, confirm/obtain the secret, switch to `./plugins/issue-duplicate-checker`, then merge.

**VERDICT:** ❌ **Needs rework** — address the failing PR-description check and the `OPENHANDS_API_KEY` confirmation before merge.

**KEY INSIGHT:** The one repo that hosts the shared action is the one place the pinned-sha indirection is actually wrong — it should dogfood the action via a local path so the action's own repo is the first to exercise every change.

> *This review was generated by an AI agent (OpenHands) on behalf of Engel.*

enyst:
Closing: taking a different route. Same two reasons as automation#409 (no `OPENHANDS_API_KEY` here, and `github.token` hits the rate bucket), plus this repo hosts the action so a pinned-sha self-reference is the wrong indirection anyway. We'll run the duplicate checker as a Cloud Automation instead of per-repo workflows.

PR #607, opened 2026-09-18
fix(iterate): ship real vendor manifest dirs for Codex install

### LINKED ISSUE #257: /iterate still fails to load in Codex because .codex-plugin symlink is lost during plugin install
## Description

`/iterate` still cannot be loaded as a Codex plugin after the previous fix in #202.

The upstream marketplace snapshot appears correct: `skills/iterate/.codex-plugin` exists and is a symlink to `.plugin`.

However, after installing the plugin with Codex CLI, the installed cache no longer contains `.codex-plugin`, so Codex still fails to load the plugin.

## Environment

- Codex CLI version: `0.132.0`
- OS: Linux x86_64
- Marketplace: `https://github.com/OpenHands/extensions.git`
- Plugin: `iterate@openhands-extensions`

## Steps to Reproduce

```bash
codex plugin marketplace upgrade openhands-extensions
codex plugin remove iterate@openhands-extensions
codex plugin add iterate@openhands-extensions
```

Then start a new Codex session or trigger plugin loading.

## Observed Behavior

The marketplace snapshot contains the expected symlink:

```text
~/.codex/.tmp/marketplaces/openhands-extensions/skills/iterate/.codex-plugin -> .plugin
```

But after installation, the plugin cache does not contain `.codex-plugin`:

```text
~/.codex/plugins/cache/openhands-extensions/iterate/1.0.0/
  .plugin/
  README.md
  SKILL.md
  commands/
  references/
```

Codex then logs:

```text
codex_core_plugins::loader: failed to load plugin: missing or invalid plugin.json plugin="iterate@openhands-extensions" path=~/.codex/plugins/cache/openhands-extensions/iterate/1.0.0
```

As a result, `/iterate` is not available in Codex.

## Expected Behavior

`iterate@openhands-extensions` should install with a valid Codex manifest path:

```text
~/.codex/plugins/cache/openhands-extensions/iterate/1.0.0/.codex-plugin/plugin.json
```

and `/iterate` should be registered as a usable command.

---
## OpenHands AI triage
**The following comments and acceptance criteria were added by the OpenHands AI agent.**

### Triage
Reproduced against `main` via the GitHub API. Every vendor directory in this registry (148 total: `skills/<name>/.codex-plugin`, `plugins/<name>/.codex-plugin`, and the `.claude-plugin` siblings) is stored as a git symlink to `.plugin` (git mode `120000`). Codex's plugin installer stages a marketplace clone with a recursive copy that only handles directories and regular files (`copy_dir_recursive` in `openai/codex` `codex-rs/core-plugins/src/store.rs`), so symlink entries are dropped. The installed cache therefore has `.plugin/` but no `.codex-plugin/`, and `find_plugin_manifest_path` (`codex-rs/utils/plugins/src/plugin_namespace.rs`) searches only `.codex-plugin/plugin.json`, `.claude-plugin/plugin.json`, and `.cursor-plugin/plugin.json` — finding none, it errors with `missing or invalid plugin.json`. The prior fix (#202) added the manifest and symlinks to the snapshot, which is why `skills/iterate/.plugin/plugin.json` exists and the marketplace snapshot looks correct, but the symlink does not survive install.

This is the same class of defect already acknowledged for the repository-root marketplace catalog in #394/PR #395, where symlinked directory paths were replaced with real directories and real files. The bounded fix is to make the vendor manifest paths real, non-symlinked files (real `.codex-plugin/` and `.claude-plugin/` directories containing `plugin.json`, with `.plugin/` remaining canonical) for plugin- and skill-directories that ship a manifest, and to update `scripts/sync_extensions.py` plus its tests so the sync/CI check enforces real files instead of symlinks. Scope is limited to the vendor manifest layout and its generator/tests; it does not change skill content, marketplace entries, or command definitions.

Non-goals: redesigning the skill/plugin packaging model; adding `.cursor-plugin` support; changing `MIGRATION.md` Python/JS catalog accessors; changing Codex or the SDK.

### Acceptance Criteria
- [ ] For every directory under `skills/` and `plugins/` that contains `.plugin/plugin.json`, `.codex-plugin/plugin.json` and `.claude-plugin/plugin.json` resolve to real regular files (git mode `100644`), not symlinks, and each contains the same JSON as `.plugin/plugin.json`.
- [ ] `.plugin/` remains the canonical hand-authored manifest location; editing `.plugin/plugin.json` once is sufficient, with vendor copies generated by `scripts/sync_extensions.py`.
- [ ] `python scripts/sync_extensions.py --check` exits 0 on the updated tree, and its `symlinks` task detects a reintroduced `.codex-plugin` symlink (or a stale/mismatched vendor manifest) as a problem and fixes it when run without `--check`.
- [ ] `skills/iterate/.plugin/plugin.json` is unchanged in content (name `iterate`, version `1.0.0`), and after install into a Codex cache the path `.codex-plugin/plugin.json` exists so `/iterate` registers.
- [ ] Existing manifest/content validations still pass (`tests/test_skill_plugin_loading.py`, `tests/test_plugin_manifest.py`, and the vendor-symlink expectations are updated to the real-file contract rather than left asserting `is_symlink()`).
- [ ] A copy of a plugin/skill directory performed by a routine that skips symlinks (e.g. `shutil.copytree` / Codex-style recursive copy) yields a target that still contains `.codex-plugin/plugin.json`, demonstrating the regression is fixed.
- [ ] No secrets or credentials are introduced, and `AGENTS.md` guidance describing vendor symlinks is updated to describe the real-file invariant so future skills follow it.
- [ ] A test or check enumerates every skill/plugin directory with a `.plugin/plugin.json` and fails when a vendor manifest path is a symlink or missing, guarding against recurrence (including for newly added skills).

onatozmenn:
HUMAN:

I ran the targeted VendorManifests tests and verified the lossy-copy case fails on base but passes with the fix, with no new failures or ruff issues beyond the existing ones.

---

## Why

`skills/iterate/.codex-plugin` is a symlink to `.plugin/`, and Codex `plugin add` drops symlinked directories when materializing its install cache. The installed tree then has no `.codex-plugin/plugin.json`, so Codex fails with "missing or invalid plugin.json" and `/iterate` never registers (#257).

## Summary

- Replace `skills/iterate/.codex-plugin` and `.claude-plugin` symlinks with real directories containing byte-identical `plugin.json` copies (mode `120000` → `100644`).
- `sync_symlinks` now accepts a real vendor dir whose `plugin.json` mirrors the canonical one, flags stale copies, and refreshes them in fix mode; missing vendors still get symlinks as before.
- New `TestVendorManifests` tests, including a lossy-install-copy simulation of the reported failure.

## Issue Number

Fixes #257

## How to Test

- `python -m pytest tests/test_sync_extensions.py -q -k VendorManifests` → 4 passed, 2 skipped (symlink-creation tests skip on Windows without privileges; they run on Linux CI).
- Lossy-copy test fails on base (manifest missing after symlink-dropping copy) and passes with the fix.
- Full `test_sync_extensions.py`: the 7 remaining failures are identical on base (Windows-env frontmatter/catalog issues), no new failures. `ruff check` on touched files: no new findings (2 pre-existing F541s on base).

## Type

Bug fix


onatozmenn:
Follow-up on the review: the branch has been cleaned to just this fix, and the valid point about plugin-loading tests is addressed — `test_all_marketplace_skills_have_vendor_manifests` and `test_has_vendor_manifests` in `tests/test_skill_plugin_loading.py` now accept a real directory mirroring `plugin.json` alongside symlinks, matching the new sync rule. The iterate cases pass; remaining failures in that file are pre-existing Windows-checkout artifacts (symlinks materialize as files), identical on base.

onatozmenn:
Fixed the Linux failure: `test_missing_vendor_still_gets_symlink` no longer asserts `sync_symlinks(check=False) == []` (fix mode returns the `missing:` entries it just fixed). It now asserts the symlinks were created and that a follow-up `check=True` run is clean.

